PR #10, opened 2026-07-15
Effect linearity: once modes and affine resumptions

jbwinters:
## Why this PR exists

Jacquard effect handlers can pause a program and later continue it. Some operations are safe to continue more than once, while others must be continued exactly once. Before this change, that rule was not part of an operation's checked definition, which made accidental double use easier to miss.

## What changed

- Each effect operation now declares a mode: `once` or `multi`.
- The type checker tracks one-use continuations and reports an error when they may be used twice.
- The interpreter and native runtime also reject double use as a safety backstop.
- The standard library now has a reviewed mode for every released operation: 13 are `once` and 7 are `multi`.
- Tests cover every one-use operation, multi-use search handlers, state handlers, and tricky branching cases.

## Why this approach

The rule belongs in the operation definition so the checker, interpreter, native compiler, documentation, and hashes all agree. Static checking catches mistakes early; runtime checks protect programs that reach execution through lower-level paths.

This PR does not add new effects or change unrelated language behavior.

## How it was checked

- 564 compiled tests and 32 command-line transcripts.
- Runtime safety checks and native/interpreter comparison.
- Native leak checks and 1,000 generated native cases.
- Release evidence reproduced from a clean overlay.

## Review and merge notes

This is the first PR in the effect-linearity stack and is based on `main`.

Please review it as a language-safety change. Do not merge it until the owner gives direct approval.


jbwinters (commented):

Built the branch and ran the full suite: 564 tests green, including the native-effects cram lane (C runtime compiled with -Wall -Wextra -Werror, interpreter/native byte-identical on the hostile cases). Reviewed the three enforcement layers against the usual attack list; the design holds:

- Per-instance semantics are right, and the paired tests prove it: `multi-around-fresh-once.jqd` (legal — each resumed branch re-performs and gets a fresh instance) vs `multi-inside-captured-once.jqd` (traps E0906 — same captured instance resumed twice). The static checker correctly passes the second case (statically that clause resumes k once) and the runtime backstop catches it, on both engines with identical output.
- Hash stability verified: `Multi` serializes as absence so legacy interfaces are byte-identical (state unchanged), and `Once` is a trailing byte so eval/abort/throw/emit change deliberately, with `diff` rendering the break as a resumption-contract change.
- The affine walk survives the escape attempts that matter: closure capture → E0817, quote capture → E0817, aliasing with shared budget → E0816, transfer into recursive/unbounded helpers → E0817. Where it can't prove correlation it over-rejects rather than under-rejects.
- Native runtime: no stale-flag hazard on pooled JQ_RESUME reuse (`jq_alloc_block` assigns flags rather than OR-ing, so USED can't survive pool recycling), flag bits don't collide, and the shared-token mechanism works because `clone_frame` shallow-dups slot values so multi-handler clones share the once block.
- Stdlib 13/7 split is defensible; State/Check/Fault staying `multi` are documented deliberate exceptions (State's transformer captures k as a value, which `once` would forbid).

Three notes, none blocking:

1. docs/effect-linearity.md has a few slop-pattern flourishes worth trimming: "Per-instance matters and deserves its paragraph." (line 41), the "unwisdom is a lint, not a mode violation" construction, "Static where checkable, dynamic everywhere", "instead of relying on folklore", and "belt-and-suspenders" twice.
2. The interpreter now does a store lookup per `perform` (src/eval.ml:687-690) that didn't exist before; native bakes `.once` in at compile time so this is interpreter-only, but worth a glance if perform-heavy interpreter loops matter.
3. An annotated handler used as a transformer (`Ann(Handle, _)` in function position) doesn't get the `Immediately_applied_transformer` context (src/check.ml:552-558 matches bare `Kernel.Handle` only), so some valid function-of-state shapes must be written in exactly the blessed form. Sound (over-rejects), and consistent with the doc's stated conservatism — flagging so it's a known sharp edge rather than a surprise.

jbwinters:
Review follow-up

Addressed in `5cdfb72`:

- Annotated immediate handlers now receive the same safe transformer treatment as bare handlers.
- Added positive and escape regressions for that case.
- Trimmed the flagged prose.

The per-perform store lookup is a performance follow-up, not a correctness problem, so it is intentionally unchanged in this release-hardening patch.

PR #11, opened 2026-07-15
Fix higher-order effect-row propagation

jbwinters:
## Why this PR exists

A higher-order function can receive another function as an argument. Jacquard was sometimes mixing up the effects used by the outer function with the effects allowed inside the callback. That could produce confusing errors or make a safe wrapper look more powerful than it really was.

## What changed

- Function application now moves effect information in the correct direction.
- A wrapper can handle an effect without that handled effect leaking back into the caller's type.
- Joining `if` or `match` branches no longer changes the callback's input requirements.
- Rigid type annotations remain strict; this is not a weakening of the checker.
- New diagnostics explain cases where a value cannot safely be generalized.
- Adversarial tests cover wrappers, callbacks, branches, and the Preflight helper.

## Why this approach

The fix is in the checker rule that connects callers and callbacks. That keeps the behavior general instead of adding special cases for one library function. It also preserves Jacquard's existing safety boundary: code still has to declare every effect it can actually perform.

## How it was checked

- 562 compiled tests plus focused command-line transcripts.
- Preflight, checker-effect, case-study, and tier tests.
- Protected public-surface manifest verification.
- Independent implementation and adversarial review.

## Review and merge notes

This PR is based directly on `main` and can be reviewed on its own.


jbwinters (commented):

Built the branch and ran the suite: 562 tests, green. This is more than a spot fix — it separates the ambient accumulator of performed effects (now a threaded `row ref` updated by immutable union) from the polymorphic tail variables that carry sharing constraints, which was the actual defect in the old `unify_rows (opened ctx frow) ambient` pattern. Checked both soundness directions:

- Effects can't escape: `include_rows` always unions the callee's fixed effects into the ambient (src/types.ml:253), lambdas land their accumulated ambient on the arrow, and the tier report shows effect counts only ever dropped (fn effectful 90→42, throw 52→14) — spurious attributions removed, nothing hidden.
- No new over-restriction: `require_effects` only specializes rows still open after argument unification, so a closed pure callback is exempt (types.ml:263), and the rigid-skolem rejection replaces an equivalent old rejection with a better message rather than shrinking the acceptance set.

Also confirmed no surviving instance of the old pattern (all four application arms route through `include_callee`), and the new adversarial tests cover the right cases: handler subtraction through a typed wrapper, branch joins in both orders, monomorphic-row sharing, E0816 with hint.

Two notes, neither blocking:

1. test/test_surface_laws.ml freezes the 0.1 release-inventory counts to constants (554/32) instead of the previous live cross-check, with the live count moved to the surface-syntax DECISION overlay. Internally consistent, but future drift in the 0.1 docs won't be caught by the suite anymore — flagging so it's a deliberate choice rather than a side effect.
2. Several demo/doc annotations were hand-loosened (e.g. stormglass `tally-policy` callback dropping `Dist`, `check.throws` callbacks dropping `Check`). Each is exactly what the fix predicts and each is pinned by a passing golden, and the wrapper's own row still retains the effect — but they're worth a skim since they're now the documented contracts.

Minor nits: `join_rows ~level:_` ignores its parameter, and `continuation_row` is included into the outer ambient twice (idempotent, harmless).

jbwinters:
Review follow-up

Addressed in `7f50352`:

- Removed the unused `join_rows` level argument.
- Clarified why continuation rows are included where they are.
- Added a regression for the row-sharing behavior.
- Updated the current live inventory while preserving the historical 0.1 freeze counts.

PR #12, opened 2026-07-15
Structured concurrency C0: pure parallel hints

jbwinters:
## Why this PR exists

The standard library needs simple building blocks that express “these pure calculations may be done in parallel.” The current runtime cannot safely run Jacquard values on multiple native threads yet, so this PR must provide the programming interface without pretending the unsafe optimization is ready.

## What changed

- Added `parallel.map` and `parallel.both` for calculations with no effects.
- Results keep a fixed, predictable order.
- Failures are reported predictably.
- The current implementation runs sequentially, so it is safe on today's runtime.
- Tests pin output hashes, ordering, failures, repeated runs, and interpreter/native agreement.
- Release evidence records stress, sanitizer, thread-safety, and benchmark results.

## Why this approach

The API expresses the programmer's intent now, while the implementation stays honest about runtime limits. A native worker pool was investigated but not enabled because Jacquard's ownership and reference-counting rules are not thread-safe yet, and failures could not be joined predictably.

The benchmark also showed that the current wrapper is slower than direct `list.map`, so enabling workers would add risk without proven benefit.

## How it was checked

- 559 compiled tests.
- Repeated stress tests, default address-sanitizer runs, ordered-failure tests, an optional/manual TSAN lane, and a 31-run benchmark.
- Interpreter/native parity and protected-manifest checks.
- Independent final review.

## Review and merge notes

This PR is based directly on `main`. It adds pure parallel *hints*, not native parallel execution.


jbwinters (commented):

Built the branch and ran the suite: 559 tests pass (exactly +5 over base), including the parity crams and property lanes. The two questions that matter for a "pure hint" design both check out:

1. Purity enforcement is airtight. The callbacks are typed with a closed empty row, and since the checker is unification-based (no pure-into-effectful subsumption), there's no path for an effectful closure to flow in. Verified adversarially beyond the in-repo tests: an inline `emit` callback, a let-bound effectful callback, and an open-tailed `() ->{emit | e}` thunk are all rejected with E0801 ("closed effect rows differ" / "cannot absorb extra effects").
2. Observable semantics are exactly sequential: `parallel.map` delegates to `list.map`; `parallel.both` forces left then right. Trap parity is tested with two distinguishable failures so source-order-first-wins is actually asserted, not assumed.

On SC.2: there are zero C changes — the native "optimization" was investigated and declined, and docs/native-parallel-decision.md grounds the no-go in real runtime constraints (exit(2) error path, non-atomic RC decrement, single-allocator freelists — all three verified against the code). That's the honest call; the doc even states explicitly that sanitizer parity is evidence for the sequential fallback, not for a worker path that doesn't exist.

Two nits: the golden prints `parallel.map` with alpha-renamed type variables relative to the doc/source form (cosmetic, internally consistent), and the PR body's "thread sanitizer" bullet could be over-read since TSAN is skipped by default — the decision doc's framing is the accurate one.

jbwinters:
Review follow-up

The implementation needed no code correction. The PR description now states plainly that AddressSanitizer is the default native check and ThreadSanitizer is optional/manual. The alpha-renamed golden spelling is cosmetic and does not change the contract.

PR #13, opened 2026-07-15
Add deterministic surface export and direct native builds

jbwinters:
## Why this PR exists

Jacquard has a friendly `.jac` source format and a lower-level `.jqd` kernel format. Native builds previously required extra manual steps, and exporting the exact lowered program was not a clear, supported workflow.

## What changed

- `jac build` can now compile a `.jac` source file directly.
- Added `jac export INPUT.jac -o OUTPUT.jqd` for debugging, conformance checks, and release evidence.
- Export uses the same lowering and name-resolution path as normal execution.
- Output is published atomically, so a failed export does not leave a partial file.
- Input is opened once and descriptor-stabilized; non-regular descriptors, destination replacement, collisions, permissions, and path-replacement failures are handled safely.
- Tests compare direct builds with exported builds across handlers, quotes, distributions, recursion, and Preflight.

## Why this approach

There should be one canonical lowering for a `.jac` program whether it is run, built directly, or exported first. Reusing the same lowering path prevents the commands from slowly drifting apart.

A successful export proves that lowering, name resolution, and canonical identity succeeded; it does not prove the program is well typed. `check`, `run`, and `build` perform typechecking.

Ordinary programs still need only the `.jac` source. Exported `.jqd` files are evidence/debug artifacts, not hand-maintained twins.

## How it was checked

- 560 compiled tests and 33 command-line transcripts.
- Full build and formatting checks.
- Member-hash and native output/error/exit-code comparisons.
- Atomic-write and filesystem race/failure tests.
- Protected-manifest verification and independent adversarial review.

## Review and merge notes

This PR is based directly on `main` and can be reviewed independently.


jbwinters (commented):

Built the branch, ran the suite (560 tests, 65s, green) plus the export cram, and reproduced the key claims by hand: exported `demos/basics/m1-fact.jac` twice from two fresh temp stores → byte-identical; `hash source.jac` equals `hash exported.jqd`; no absolute paths or dates leak into output; a second export to an occupied path fails with E1301 and leaves the existing file untouched.

The safety question I cared most about checks out: the direct native build does not bypass validation. `build_cmd` still runs `Check.check_top` on every top (bin/main.ml:1742,1746) and `Build.build` still enforces refusals and bakes the capability manifest. An ill-typed `.jqd` is rejected with E0801 at build time. The atomic-publish logic in src/export.ml (O_EXCL temp + `Unix.link`, fsync, full rollback, orphan diagnosis) held up under the fault-injection tests and my collision experiment.

Three nits, none blocking:

1. `export` itself does not typecheck — the path is resolve → `Canon.hash_top` → print (bin/main.ml:1689-1699). `printf 'add(1, True)' | export` succeeds; `run`/`build` on the result correctly reject with E0801. Defensible (mirrors `hash`, and nothing unchecked can execute), but a successful export is not evidence of well-typedness; worth one doc line so nobody reads it that way.
2. The PR body says symlinks are refused; they aren't — export.ml:45 opens without O_NOFOLLOW and the project's own test expects a symlink input to be read (test/test_export.ml:51-52). What's actually enforced is descriptor stability plus rejection of non-regular descriptors, and the write side can't follow a destination symlink. Safe, but the body's wording overstates it.
3. `build`/`export` resolve all tops before checking (bin/main.ml:85-107, then 1737/1752) while `check`/`run` interleave per top, so a program with an early type error and a late resolution error can surface diagnostics in a different order per command. Diagnostics-only; no artifact is produced either way.

jbwinters:
Review follow-up

Addressed in `9dad0e7` and the PR description:

- Documentation now says export resolves, canonicalizes, and publishes syntax but does not typecheck it.
- The symlink claim now describes descriptor stability and destination safety accurately.

`run`, `check`, and `build` remain the commands that typecheck before execution or artifact publication. Cross-command diagnostic ordering is a diagnostics-only follow-up.

PR #14, opened 2026-07-15
Effect taxonomy: blessed identities and risk metadata

jbwinters:
## Why this PR exists

Jacquard needs a single trusted list of effects such as Filesystem, Network, Audit, and Secret. A name alone is not enough: an untrusted program could create a different effect with the same display name. Review tools also need useful risk information instead of treating every effect as identical.

## What changed

- Defined the complete 25-slot effect catalog: 12 released effects and 13 reserved slots.
- Added a registry keyed by the effect's exact hash, not just its name.
- Recorded each effect's tier, default risk, meaning, namespace, interface rules, and operation schemas.
- Added readable authority differences for tools that compare what two programs may do.
- Native grant handling now preserves exact identities, so a look-alike name is never treated as an official effect.
- Documented the future Async child-effect rule without claiming that the later concurrency runtime already exists.

## Why this approach

Exact hashes let Jacquard tell the difference between the official Network effect and an imitation also called “Network.” The registry gives the checker, CLI, native runtime, documentation, and future review tools one shared source of truth.

Reserved slots make future planning visible without assigning a fake identity before an interface is actually released.

## How it was checked

- 578 compiled tests and 22 documentation examples.
- Native/interpreter comparison: 69 identical programs, 8 expected refusals, 0 failures.
- Exact-identity spoof tests and both protected release manifests.
- Independent native-identity review.

## Review and merge notes

This PR is stacked on PR #10 because it uses the operation-mode work introduced there.

Review PR #10 first. Do not merge this PR until the owner gives direct approval.


jbwinters (commented):

Built the branch and ran the suite: 578 tests green (86s), including the new effect-taxonomy suite and the spoof crams; also reproduced the spoof and positive-control behavior by hand from the worktree binary.

The central claim — identity anchored in the interface hash with no name-based fallback — holds on every path that matters:

- Checker: `Effect_registry.find_canonical` keys on the row's hash (src/check.ml:1494-1496); the name is display-only. A user effect always renders as `unrated user effect`.
- Interpreter grants: `granted_hashes` is snapshotted against the trusted prelude before user code loads (bin/main.ml:170-174 vs 215), and the manifest check is pure `Hash.equal`. A program that shadows `console` can't pollute the snapshot — verified empirically: spoofed console refuses identically on both engines (exit 3, unrated), legit console runs under `--allow console`.
- Native: effect flags are keyed by hash hex (`g_eff_<hash>`), native ops install only on `Hash.equal` of the owning DefEffect hash, and the eval refusal is keyed by canonical identity + index name, not the effect's display string.

Reserved slots are genuinely inert (no hash, `register` refuses them, `canonical` folds them out, retagging rejected), and risk metadata feeds only rendering — grant gating uses `index_name` membership, so risk can't silently change semantics.

One note and one nit, neither blocking:

1. The 25-row catalog now lives in three hand-edited copies (spec TSV, docs table, OCaml `catalog` in src/effect_registry.ml:1039-1096). Drift is test-enforced in all three directions (TSV↔doc, TSV↔registry, registry↔prelude), so it can't diverge silently, but it's a standing maintenance tax — if it grows past 25 slots it may be worth generating two of the three from one source.
2. `show_row` collision detection is O(n²) over the effects in a single row (src/check.ml:1461-1473) — immaterial at row sizes, noting for completeness.

jbwinters:
Review follow-up

No correctness change was needed. The catalog duplication and row-rendering complexity are documented maintenance/performance costs, and the existing drift tests keep the three catalog copies synchronized for the 0.1 release.

PR #15, opened 2026-07-15
Structured concurrency: freeze Task and Async interfaces

jbwinters:
## Why this PR exists

Before adding a scheduler, Jacquard needs a precise contract for tasks and async scopes. Without that contract, later runtime work could accidentally change task identity, cancellation, waiting order, or which child effects are charged to the parent.

## What changed

- Defined the public `Task`, `TaskResult`, and `Async` interfaces.
- Specified task creation, completion, failure, waiting, cancellation, ordering, deadlock, and escape behavior.
- Pinned the exact interface hashes and reject look-alike names or schemas.
- Added direct checking so a spawned child's effects are visible to the caller.
- Added a safe hash encoding for a self-referencing effect row while preserving all previously valid hashes.
- Added executable fixtures and complete SC.0 release evidence.

## Why this approach

This PR freezes the rules before introducing task values or a scheduler. That makes later implementation PRs measurable: they must match these contracts instead of inventing behavior as they go.

The scope is deliberately narrow. It does not add host threads, task values, or runtime scheduling. Higher-order wrappers are handled in a later PR.

## How it was checked

- 568 compiled tests.
- 24 documentation examples across 7 documents.
- Exact interface/hash checks and structured-concurrency manifests.
- Independent contract and identity review.

## Review and merge notes

This PR is stacked on PR #14 because Async is part of the frozen effect catalog.

Review PRs #10 and #14 first. Do not merge this PR until the owner gives direct approval.


jbwinters (commented):
**Sign-off: LGTM, approve-with-nits — one factual doc error to fix before merge.** (Posted as a comment — GitHub blocks approving from the author's own account.)

Built the branch and ran the suite: 568 tests green (95s); the 24-doctest claim checks out exactly. The part of this PR I treated as highest-risk — touching canonical hashing to encode the self-referencing Async row — is sound:

- The `0x38` row tag is emitted only when a `Self` entry is present; the else branch reproduces the old `0x36` byte stream exactly, including sort order (src/canon.ml:275-311). `effectself` is set only inside effect serialization (src/canon.ml:524) and defaults to `None` everywhere else, so a DefType row can never emit `0x38`.
- "Preserves all previously valid hashes" is genuinely enforced, not aspirational: the old `the_hash` hard-errored (E0501) on any Named row member, so no previously valid interface could hit the new path. Corroborated by the pinned legacy Abort hash and the byte-level assertion that its payload contains no `0x38`.
- Child-row charging is scoped to the exact resolved `async.spawn` identity (src/check.ml:687-696), and the doctest pins the observable contract: `spawn-net : () ->{Async, Net} Task Text`, `scoped-net : () ->{Net} ...` — scope discharges Async, preserves Net, per the SC.0 requirement. All task_124 exclusions are recorded in §6.

Three findings, first one should be fixed (it's a one-word edit, no code):

1. docs/release/effect-linearity/EVIDENCE.md:235 says "34 cram transcripts" but the repo has exactly 32 `.t` files (none added here), and structured-concurrency/EVIDENCE.md:100 in this same PR correctly says 32. The two evidence docs contradict each other; 32 is right.
2. concurrency.md:150-152 declares the deadlock message templates "exact", but they exist only in prose — `grep -rn deadlock src/ test/` is empty. The escape diagnostic got the `task_escape_code`/`task_escape_message` treatment; the deadlock templates should get the same if a later scheduler PR is to be held to them.
3. `cancellation_points = [Await; Yield; Routed_effect]` encodes the doc's "exact contract order" but the only test asserts length 3, not sequence (test_effect_taxonomy.ml:857) — a reordering would pass the suite.

jbwinters:
Review follow-up

Addressed in `1009066` and `ba20797`:

- Corrected the evidence count from 34 cram transcripts to 32.
- Pinned the exact cancellation-point order.
- Moved the self-await and wait-cycle messages into compiled contract helpers and asserted their exact text.

PR #16, opened 2026-07-15
Improve nested delimiter recovery diagnostics

jbwinters:
## Why this PR exists

When a source file is missing a closing brace, parenthesis, or other delimiter, the parser used to produce a cascade of follow-up errors. That made one small typo look like many unrelated problems and could hide useful errors later in the file.

## What changed

- Added structure-aware recovery for quotes, matches, `if`, handlers, and blocks.
- Diagnostics now point to both the opening delimiter and the place where parsing failed.
- The checker can continue far enough to report safe errors in later top-level definitions.
- Every recovered syntax tree remains marked invalid.
- Strict commands such as format, hash, and run still refuse malformed files.
- Added realistic broken examples and 300 generated delimiter mutations.

## Why this approach

A checking command should help the programmer find more than the first typo, but recovery must never turn broken code into runnable code. The parser therefore recovers only for diagnostics, while every command that publishes or executes meaning stays strict.

## How it was checked

- 554 compiled tests plus exact malformed-file transcripts.
- Format/check/hash/run refusal tests with exact nonzero exits.
- Valid-file hash and operation-namespace stability checks.
- Protected surface manifest and independent adversarial review.

## Review and merge notes

This PR is based directly on `main` and can be reviewed independently.


jbwinters (commented):

Reviewed the full diff against main, built the branch, and ran the suite (554 tests pass, including the new exact-diagnostic goldens, the bounded-cascade check, and the 300-case seeded mutation property).

The recovery design is right: primary span at the failure token, hint naming the opener's span, one E1220/E1221 per malformed file, and recovery confined to `check` while fmt/hash/run stay strict. I looked specifically for valid-program regressions and found none — `top_item_ahead` only matches tokens that are never legal block contents (src/surface_parse.ml:318-328), no `corpus/valid/*` hash changed, and the new format/hash idempotence test covers the valid nested files.

Three nits, none blocking:

1. src/surface_parse.ml:1439 — recovered blocks set only `surface-form` ("recovery-delimiter") instead of going through `mark_recovered_construct`, so they lack the `surface_hole` ID every other recovered construct gets. Not a functional gap (the pushed `expr_hole` item still trips `has_holes_expr`, and the strict-rejection test passes), but it diverges from the pattern and from the docs/errors.md:172-176 claim that every marker carries a stable surface-hole ID.
2. The quote message "expected `}` before the enclosing boundary" is vaguer than its siblings ("before the next top-level item", "found `]`"). Understandable since the failure is a surrounding delimiter, but worth a second look.
3. docs/SKILL.md:387-391 and docs/surface-syntax.md:653-659 add quote-authoring guidance that's beyond this PR's scope.

jbwinters:
Review follow-up

Addressed in `12528ee` and `8252187`:

- Recovered blocks now carry stable surface-hole IDs like other recovered constructs.
- The quote-boundary diagnostic is clearer.
- Unrelated quote-authoring guidance was removed.
- The nested-brace regression now pins the new deterministic hole-ID order; this also fixes the clean-runner CI failure.

PR #17, opened 2026-07-15
SC.3: add scoped opaque task values

jbwinters:
## Why this PR exists

The Async interface was defined in PR #15, but Jacquard still needed a real task value that the runtime could safely carry. A task from one run or scope must not be usable in another run, stored forever, forged, or confused with an ordinary value.

## What changed

- Added opaque task values with run-local and scope-local ownership.
- Enforced task ownership in the evaluator, cache, continuations, one-use resumes, store, checker, and native runtime.
- Aligned task-ID limits between OCaml and native code.
- Malformed, private, escaped, or cross-run task values fail closed.
- Added runtime and native tests for ownership, cleanup, and forged values.

## Why this approach

Tasks are capabilities: holding one gives access to a specific running computation. Keeping the representation private and checking ownership at every boundary prevents a task handle from escaping into a place where it no longer has a valid meaning.

This PR adds task values only. Scheduling behavior arrives in later stacked PRs.

## How it was checked

- 572 compiled/property tests and 33 command-line transcripts.
- Focused address/undefined-behavior sanitizer runs.
- Interpreter/native parity and structured-concurrency manifest checks.
- Independent final review.

## Review and merge notes

This PR is stacked on PR #15, which defines the public Task and Async contracts.

Review the stack through PR #15 first. Do not merge this PR until the owner gives direct approval.


jbwinters (commented):

Built the branch, ran the suite (572 tests green, +4 over base), and the ASan runtime check passes including the new inert-task-handle cases. The properties that matter at this milestone all verify:

- Opacity is real: no public constructor (the sole OCaml seam is unexported and uncalled; native creation refused at lowering), no structural equality (physical only, constant fingerprint), `<task>` redaction in both engines, and the frozen private-hash rejection is enforced at all four tiers — checker, interpreter, native lowerer, and store. Best evidence it's hash-derived rather than name-gated: a structurally identical local `type Task a = | TaskOpaque` collides with the frozen private hash and is rejected with E0907.
- Ownership validation fails closed (E0907, never a crash) at every evaluator boundary, including captured continuations, and `Once_state` binds an owner run token so cross-context resume is refused before the affine budget is consumed — good composition with the linearity stack (once-hostile generator now 17/17, both engines).
- spawn charging inherited intact: the doctest still pins `spawn-net : () ->{Async, Net} Task Text` / `scoped-net : () ->{Net} ...`.

Notes, none blocking:

1. src/eval.ml:193 `scheduler_task_value` — the only genuine OCaml VTask mint — is warning-suppressed, unexported, and has zero live coverage (all OCaml tests forge via Obj.magic). Fine as a forward seam for SC.5, but it means the real construction/validation round-trip is untested on the OCaml side until then.
2. The Obj.magic forgeries in test/test_prelude.ml:1807-1810 depend on exact record field order for both `ctx` and `Task_handle.t`; a reordering would silently forge wrong data and could flip these ownership tests to false passes. A comment pinning the layout dependency (or a constructor-based test seam in SC.5) would make this less fragile.
3. Expected scope note: since no scope executes yet, the end-to-end escape scenarios are structurally unexercisable here — the Obj.magic divergent-scope-path forgeries are the right stand-in, and real escape coverage lands with SC.6.

jbwinters:
Review follow-up

Addressed in `50b9b3c`: the forged-task tests now document both record-layout assumptions behind their `Obj.magic` fixtures. The live mint/validation round trip and real escape coverage arrive in the later scheduler integration PR, where they are tested end to end.

PR #18, opened 2026-07-15
GM.0: freeze governed membrane charter

jbwinters:
## Why this PR exists

Jacquard is adding “governed membranes”: a review layer between an agent's request and powerful actions such as reading files or using the network. Before implementing handlers, the project needs one written contract for what is reviewed, what is logged, and what authority can cross the boundary.

## What changed

- Froze decisions D61–D73 as the first governed-membrane charter.
- Defined how a call keeps its original operation, arguments, authority, policy, assessment, proposal, and audit history.
- Required one audit sequence for the whole run.
- Separated live execution from dry simulation.
- Required secrets to appear only as safe references, not raw values.
- Defined nesting and Eval boundaries.
- Added checker-backed example signatures for agent, live workspace, and dry-run workspace code.

## Why this approach

Security-sensitive code should be built from a reviewed contract, not from handler behavior that is explained afterward. This PR creates that contract and executable type evidence while deliberately avoiding implementation.

The one-row-tail design keeps the effect types readable and makes hidden authority harder to smuggle through helper functions.

## How it was checked

- Direct fixture output matches the expected signatures.
- 23 documentation examples.
- Focused effect-taxonomy checks.
- Full build and formatting checks.
- Independent final review.

## Review and merge notes

This PR is stacked on PR #14 because the charter refers to the frozen effect catalog.

It is a contract/documentation boundary, not a live workspace implementation. Do not merge it until the owner gives direct approval.


jbwinters (commented):

Built the branch, ran the suite (568 green, doctests now 23 examples from 7 documents, membrane fixture checked against exact expected signatures). Reviewed this as a security design doc, since that's what it is. The load-bearing decisions hold up:

- Call/proposal identity split: proposal-id hashes (GovernanceV0, call-id, policy-id, assessment-id, authority, preview, rendering, summary), and the membrane rejects any decision whose embedded proposal-id fails recomputation — so an approval for call A can't be replayed against call B. Hash inputs and exclusions are enumerated precisely, and the world-state TOCTOU gap is honestly disclosed and punted to preconditions + driver.
- Audit ordering: single owner (`governance.with-sequence`), zero-start, per-entry increment, duplicate/skip/decrease rejected, pre-action writes fail-closed, and refusal paths are covered by the pre-action Evaluated entry — every executable path has a pre-action record.
- Simulation law: dry-run safety is structural, not disciplinary — the checked dry row is `{Judge, Audit}` and simulators take empty-row thunks. No operation exists in both live and dry APIs; missing simulator refuses with NoSimulation, never falls back to live.
- Eval prohibition and monotone nesting are stated absolutely (rejected even if an inner handler appears to discharge Eval; inner allow can't force outer allow), with parent-call-id lineage.
- Definition of done met: no TBD markers, all of D61-D73 indexed, Tool/Host sketch marked superseded.

Three findings, the first worth an edit before merge since this doc is normative:

1. §12.2 (line ~1160) requires every clause to "record completion before consuming its local Resume", but the reference refusal branches (lines 535, 558, 582, 1109) deliberately consume Resume with only the gate's Evaluated entry, no Completed record. Taken literally, §12.2 contradicts §7/§8, and a G2 implementer reconciling them could go either way. Scope it to action/simulation paths.
2. Know what the executable evidence actually pins: the live/dry row split and State discharge — not the disposition-gate API or the audit-position claims, which remain prose (the charter discloses this at ~729-731, so it's a scope note, not a defect).
3. Slop-pattern lines to trim in a normative doc: "That restraint is the strongest part of the design: the membrane is impressive because..." (line 1396), "That pair of signatures is the design in miniature." (36), and the "This is not a policy engine bolted beside the language. It is..." construction (38).

jbwinters:
Review follow-up

Addressed in `c6ea211`:

- Charter section 12.2 now requires `Completed` before Resume consumption only on action and simulation paths.
- Refusal paths are explicitly allowed to consume after the pre-action `Evaluated` record without inventing a fake completion.
- The evidence scope now distinguishes executable row/State checks from later gate/audit obligations.
- Trimmed the flagged prose.

PR #19, opened 2026-07-15
SC.4: charge spawned child effects

jbwinters:
## Why this PR exists

A parent must be charged for every effect its child task can perform. Otherwise a helper could hide Network, Filesystem, or another effect inside `async.spawn`, making the parent look safer than the program really is.

## What changed

- Made the exact frozen `async.spawn` operation carry the child function’s effect row into the caller’s effect row.
- Kept child effects visible through aliases, higher-order wrappers, returned closures, tuples, polymorphic rows, and nested scopes.
- Made `async.scope` remove only Async; it does not erase Network, Filesystem, or another child authority.
- Scoped this special rule to the complete official Async identity, so a look-alike user effect does not receive privileged behavior.
- Replaced an internal impossible-case assertion with fail-closed E0805 if frozen recognition and declaration conversion ever disagree.
- Added diagnostics and positive, negative, native, and look-alike tests.
- Reconciled the change with current main and the current evidence checker.

## Why it was done this way

Authority should follow behavior. If a child may use the network, the surrounding program must also admit that network authority even when `spawn` is hidden behind a helper.

The rule checks the whole frozen Async declaration instead of a name. That prevents unrelated operations with similar spelling from receiving special treatment.

## How it was checked

- Full forced suite: 669 compiled/property tests passed.
- Laundering attempts through aliases, wrappers, returned closures, tuples, row annotations, and cyclic rows are pinned.
- Pure children remain exactly `{Async}`; children using world effects keep those effects visible.
- Build, formatting, generated documentation, native parity, all 199 structured manifest entries, and the complete-overlay checker passed.

## Review and merge notes

This PR is now based directly on `main`. It changes the static rule for the exact official Async identity; it does not add the root scheduler.


jbwinters (commented):

Built the branch: 573 tests green (135s), doctests at 24 examples. This closes SC.0's higher-order hole the right way — deleting the syntactic direct-call bridge and baking the law into spawn's type: `op_scheme` makes the operation's callable row and the child-thunk's row the same physical row record (src/check.ml:412-419), so the charge is carried by instantiation through the ordinary application path, not by pattern-matching call sites.

Verified against every laundering shape worth trying, all pinned in fixtures/crams: alias (`spawn-alias`), higher-order wrapper (`forward(spawner, child)`), returned closure, tuple storage — all show `{Async, <child>}` in the checked signature. The two adversarial negatives are the convincing part: the row-polymorphic shape that tries to make callable-row ≠ thunk-row is unsatisfiable (E0801 occurs check, test/cli/manifest.t:655), and the closed `{Async}` annotation through an alias is rejected with E0804 showing the propagated Net. Also confirmed no over-charging (pure thunk yields exactly `{Async}`; fresh `e` per reference) and that the fixture handler drops the thunk, so Net in the signature is the SC.4 rule doing real work rather than a call-site charge. Old `async_spawn_child_row` is fully gone.

Two notes, neither blocking:

1. Be explicit that the non-laundering guarantee is scoped to the exact frozen Async identity: any user-declared effect whose operation takes a thunk gets the ordinary closed row, so its thunk's latent effects are charged where the thunk runs (inside the handler), not to the lexical caller. That's standard higher-order effect semantics and the docs don't overclaim it — but "no effect can hide inside a task" should always be quoted with "for the blessed Async" attached.
2. src/check.ml:416 has `| _ -> assert false` on the frozen path — reachable only if `is_frozen_async_spawn` and `conv_decl_ty` ever disagree about the pinned shape, so effectively dead, but a defensive E0805 would fail closed instead of crashing if a future refactor breaks that agreement.

jbwinters:
Review follow-up

Addressed in `564d570`:

- The non-laundering claim is explicitly scoped to the exact frozen Async identity.
- A future mismatch between frozen recognition and declaration conversion now fails closed with E0805 instead of reaching `assert false`.
- Added regression coverage and updated the evidence manifest.

PR #20, opened 2026-07-15
SC.5: add scheduler task state machine

jbwinters:
## Why this PR exists

Task values need a small, predictable state machine before a real scheduler can use them. The project must agree on task IDs, waiting order, cancellation, failures, cleanup, and dependency cycles without mixing those rules with one scheduling policy.

## What changed

- Added a policy-independent scheduler core with explicit task states and deterministic IDs.
- Defined stable waiting and wake-up order.
- Defined success, failure, cancellation requests, cleanup, and atomic dependency-cycle handling.
- Kept every saved continuation owned by exactly one task and allowed checkout only by its current owner.
- Added the full illegal-transition rejection matrix.
- Added a property proving the implementation agrees with the frozen `valid_transition` relation.
- Pinned back-to-back deterministic runs and cross-member cycle wake order.
- Documented the v0 cancellation choice: a suspended continuation is dropped rather than resumed to unwind.
- Reconciled the change with current main and the current evidence checker.

## Why it was done this way

Separating task lifecycle rules from round-robin or another scheduling policy makes ownership and cleanup easier to inspect. A later scheduler can choose which runnable task goes next without changing what the task states mean.

Atomic cycle handling matters because a detected deadlock must not leave half the cycle terminal and the other half runnable.

## How it was checked

- Full forced suite: 668 compiled/property tests passed.
- Two 200-case scheduler properties and the complete illegal-transition matrix passed.
- Deterministic IDs, waiting order, cycle teardown, foreign handles, duplicate checkout, and terminal immutability are pinned.
- Build, formatting, generated documentation, all 197 structured manifest entries, and the complete-overlay checker passed.

## Review and merge notes

This PR is now based directly on `main`. It provides scheduler machinery, not the final Async root handler or queue policy.


jbwinters (commented):

Built the branch: 572 tests green (131s), both 200-case QCheck properties included. The lifecycle core is faithful to the SC.0 freeze: every producible transition lies in the frozen legal set (terminal immutability guarded, couldn't construct an illegal transition through the public API), determinism is structural (ordered lists, monotonic spawn counter, no clock, no hash-order iteration), and the deadlock templates are reproduced verbatim and asserted against literal strings — closing the gap I flagged on the SC.0 PR where they were prose-only. The atomic cycle teardown terminalizes every member before any wakeup, and that's property-tested. Once-resume composition is sound: at most one live resume token per task, never handed out twice. Native is honestly deferred; the module is polymorphic over the resume token precisely so a native continuation can reuse it.

Findings, in descending order of how much I'd want them addressed:

1. Illegal-transition coverage is partial. Duplicate checkout, double-terminalize, spawn-into-closed, and foreign-handle are tested, but several rejection branches the code carries are unexercised: `suspend_yield` on Suspended, `wake_yielded` on Runnable (scheduler_core.ml:157), `checkout` of Suspended/terminal (:124-129), `complete` of Suspended, `await` with a not-checked-out waiter (:131-139). The DoD asks for illegal-transition rejection coverage; these are cheap tests to add.
2. Design choice to confirm explicitly: `terminalize`/`fail_cycle`/`deliver_cancel` drop a suspended task's owned continuation (entry.resume <- None) rather than resuming-to-unwind, while `close` collects tokens for explicit destruction. docs/concurrency.md documents this, so it's intentional — but it means cancelled/deadlocked tasks never run cleanup. If v0's contract is "drop, don't unwind," say so where SC.7 will read it, because cancellation delivery is where this becomes observable.
3. `Concurrency_contract.valid_transition` is exposed as the authoritative legality predicate but scheduler_core never calls it — legality is re-derived in ad-hoc matches. Consistent today; a one-line property test asserting agreement would prevent drift.
4. Minor: no back-to-back identical-schedule test (structurally guaranteed, cheap to pin), and cross-member cycle wakeup order (discovery-order grouped, then registration order) is deterministic but neither frozen nor tested.

jbwinters:
Review follow-up

Addressed in `d863180`:

- Added the missing illegal-transition rejection matrix.
- Added agreement coverage for the frozen `valid_transition` relation.
- Pinned deterministic back-to-back runs and cross-member cycle wake order.

The v0 drop-don't-unwind cancellation decision is carried forward and made type-enforced in PR #24.

PR #21, opened 2026-07-15
SC.6: add structured task scopes

jbwinters:
## Why this PR exists

Structured concurrency means every child task belongs to a clear scope. When that scope ends, all of its children and saved continuations must be finished or cleaned up. Without that rule, hidden work could survive after its parent returns and leak into later runs.

## What changed

- Added nested task scopes inside one scheduler run.
- Registered each child task with the scope that owns it.
- Cleaned up child scopes, unfinished tasks, waiting edges, and saved continuations recursively.
- Rejected task handles that escape the scope where they were created, including handles hidden inside tuples, constructors, closures, cyclic environments, and resumptions.
- Made normal returns, language errors, and host exceptions use the same cleanup bracket.
- Kept repeated close operations safe and idempotent.
- Combined the scheduler state machine from #20 with the child-effect charging rule from #19.

## Why it was done this way

A scope gives every child one lifetime and one owner. Central registration and recursive cleanup turn “no work survives the scope” into a runtime guarantee instead of a rule that every library author must remember.

The escape check walks the complete reachable value graph because hiding a task handle inside another value must not bypass the lifetime rule.

## How it was checked

- A 200-case cleanup property proves nested scopes return every ownership counter to zero.
- Tests cover joins, forgotten children, nested lineage, returned and stored escapes, foreign handles, stale operations, language errors, and host exceptions.
- Native sanitizer stress tears down 4,096 task and continuation-shaped values at a scope boundary.
- Build, formatting, generated documentation, all 55 structured overlay hashes, and the complete-overlay checker passed.

## Review and merge notes

This PR is based directly on `main`; #19 and #20 are now merged. It adds scope ownership and cleanup, not the runnable queue, cancellation routing, failure policy, or final Async root handler.


jbwinters (commented):
**Sign-off: LGTM as a library-layer milestone, approve-with-nits — but read the scope caveat.** (Posted as a comment — GitHub blocks approving from the author's own account.)

Built the branch: 573 tests green (121s). The three modules are correct in isolation and well-tested: `protect` closes on Ok/Error/exception via recursive descendant close with metrics returning to zero (200-case QCheck property), scope paths are deterministic one-based hierarchies, the escape walk covers every Value.t and frame variant, terminates on cyclic environments via physical-identity memoization, and uses the frozen E0907 code/message. The Async-discharge/Net-preserve law is correctly inherited from SC.4 (no checker changes here) and its fixtures still pass.

The caveat that should be explicit in the merge decision: this PR's headline capability is not end-to-end. `reject_task_escape`, `Structured_scope`, and `Scheduler_core` are referenced only from the two new test files; `scheduler_task_value` remains the inert, warning-suppressed seam from SC.3; there is no async.scope execution path or evaluator return-site check. Because running code can't mint a Task, the escape tests forge them with `Obj.new_block` — they validate the walk's shape, not that a genuine scoped Task is rejected on a genuine return. The docs are honest about this (concurrency.md and EVIDENCE.md both state the deferral), so the question is only whether SC.6's bar was "library layer" or "escape becomes real." If the latter, that bar moves to the SC.9 integration, and I'm holding that PR to it.

Smaller findings:

1. `Scheduler_core.close` reclaims only tokens still held by entries — a checked-out (in-flight) token is invisible to it, so `protect`'s exception path can't drop one. Latent until a real run loop checks tokens out; worth pinning before SC.9.
2. `reject_task_escape` treats `VCode` as a leaf, so a Task inside quoted Form would be missed — likely unreachable given Task opacity, noting for the "full reachable graph" claim.
3. Nits: `close ~reason` ignores its parameter entirely; `scope_prefix` (eval.ml:209) and `is_prefix` (structured_scope.ml:70) are duplicates.

jbwinters:
Review follow-up

The scope caveat is closed by the later scheduler integration: real code mints real Task values and the escape guard runs end to end. The checked-out-token cleanup gap is fixed in PR #30 with a checkout bracket that restores unsettled ownership on normal return, diagnostic error, and host exception. VCode remains syntax data and cannot contain the private runtime Task carrier.

PR #22, opened 2026-07-15
GM.1: add governed membrane core types

jbwinters:
## Why this PR exists

The governed-membrane charter in PR #18 describes what a reviewable tool call must contain. This PR turns that contract into real Jacquard data types with safe constructors and validators.

## What changed

- Added versioned governance values for calls, authority, assessments, proposals, outcomes, and related records.
- A governed call is tied to the exact effect operation found in the Store, not a display name supplied by the caller.
- Constructors return explicit `Result` errors for invalid input.
- Validators check authority order, canonical hashes, and relationships between records.
- The values live in ring 3, matching their intended governance role.
- Integrated the already-reviewed approval identity dependency.

## Why this approach

Review and audit records are only useful if they cannot quietly change meaning. Exact operation resolution and canonical hashes make two equal records mean the same thing across the checker, evaluator, native runtime, and evidence tools.

Safe constructors keep ordinary invalid data out without relying on exceptions.

## How it was checked

- 610 compiled/property tests and 32 command-line transcripts.
- Focused governance, approval, taxonomy, ring, and property suites.
- Full build, formatting, documentation, and tier checks.
- Exact GM.1 release manifest.
- Independent final review.

## Review and merge notes

This PR is stacked on PR #18 and includes the validated approval dependency from the effect-taxonomy work.

Review the charter first. Do not merge this PR until the owner gives direct approval.


jbwinters (commented):
**Sign-off: LGTM, approve-with-nits — one deliberate seam to confirm you own.** (Posted as a comment — GitHub blocks approving from the author's own account.)

Built the branch: 610 tests green (141s), fmt clean. Verified the types against the charter field-by-field (§4/§5/§6): Call's 9 fields in order, Assessment/OutcomeSummary versioned, policy shapes and ToolError constructors match, verdict laws implement Forbidden absorption and under-confidence-never-Allow correctly, and the malformed operation-name corpus traces through `operation-name-loop` case-by-case. Call identity is Store-derived (caller supplies only a name; the native resolves the exact operation hash) and hashed over the versioned semantic subject excluding display fields, per §10. SecretRef discipline is structural — no governance type has a Secret-typed field and the Show functions render only enum labels and trusted text.

The finding to consciously accept, not a blocker:

1. The charter freezes ONE versioned AuditEntry carrying the versioned Assessment with `version`/`sequence` fields; the implementation intentionally keeps the ET.2/ET.6 unversioned carriers for the audit stream and adds the GM.1 versioned family alongside (prelude/19-audit.jqd:10-49 vs 21-governance-core.jqd:37-50 — the file-top comment declares this deliberate). Reconciliation is deferred to the gate PRs. Fine as staging, but decide now which Assessment flows into audit entries and where version/sequence land, so GM.6/GM.7 don't inherit the seam silently.

Also worth knowing (by design, matching §10): the v0 constructors are public and forgeable — `validate-call`/`validate-bound-*` recompute and reject forged hashes (tested), but nothing auto-invokes the validators yet. Until the gate lands, any downstream consumer that forgets to validate accepts forged carriers. That's the charter's verifier-backstop model, just flagging the interim exposure window.

Nits: `tool-error` names every payload field `detail` where the charter says `reason`/`message` (cosmetic; ToolError never enters an identity hash); the "no generic Secret renderer" test assertion at test_governance_core.ml:313-320 has two near-vacuous conjuncts that make it read stronger than it is.

jbwinters:
Review follow-up

The versioned/unversioned carrier seam is resolved by PR #32: the dry gate writes the versioned governance assessment/outcome family into its versioned audit entries. The earlier ET audit stream remains a separate historical carrier. Live gate work must keep validating every public, forgeable carrier at its boundary.

PR #23, opened 2026-07-15
GM.2: add canonical governed identities

jbwinters:
## Why this PR exists

A governed call or approval proposal needs a stable identity. Review metadata such as a nicer summary must not change that identity, while meaningful changes such as a different operation, argument, policy, or assessment must change it.

## What changed

- Added versioned canonical encodings for `GovernanceCall` and `GovernanceProposal`.
- Call identity includes the resolved operation, arguments, authority, and preconditions.
- Proposal identity also includes the policy, assessment, preview, rendering, and outcome summary fields required by the frozen contract.
- Human-facing metadata is kept separate where the contract says it should not affect identity.
- Reused the existing `code.hash` mechanism instead of introducing a second hashing system.
- Added generated tests for stable and changing identities.

## Why this approach

Content-based identities make review records portable and tamper-evident. They also reduce sensitivity to renames and formatting changes while still changing when the program's governed meaning changes.

Using one canonical encoding prevents different parts of the system from disagreeing about what was approved.

## How it was checked

- 614 compiled tests, including 460 generated governance cases.
- Focused governance, approval, taxonomy, ring, and tier checks.
- Full build, formatting, documentation, and golden checks.
- Exact GM.2 manifest.
- Independent final review.

## Review and merge notes

This PR is stacked on PR #22, which introduces the governance data types.

Review PR #22 first. Do not merge this PR until the owner gives direct approval.


jbwinters (commented):

Built the branch: 614 tests green (134s), governance-core 14/14 including the new proposal-hash case and the three new QCheck properties, tiers cram and fmt clean. Traced the hash path end to end — `code.hash` is SHA-256/HASH_V0 over the compact printer serialization on both engines — and the two security properties I cared about hold:

- Collision-free by construction: fields combine via `code.form`, serializing as parenthesized s-expressions with self-delimiting, escaped nodes. The "ab"+"c" vs "a"+"bc" ambiguity can't arise; text can't inject structural delimiters; metadata is erased from the compact bytes so identity is formatting-stable.
- The hashed tuple matches charter §10 byte-for-byte in order: (GovernanceV0, call-id, policy-id, assessment-id, authority, preview, rendering, summary), carried proposal-id excluded. Call-id correctly excludes operation-name and summary. Field-sensitivity is QCheck-pinned across all 4 call fields and all 7 proposal fields.

The note that matters for the gate PRs: verification is two-tier. `validate-proposal` proves only internal consistency (carried hash matches own canonical bytes); anti-substitution — that the embedded call/policy/assessment ids correspond to the real artifacts, with byte-for-byte authority compare — requires `validate-proposal-artifacts`. The split is documented and intentional, but the membrane's decision boundary must call the artifacts-aware check. I'll hold GM.6 to that.

Nits: `make-proposal` validates the assessment then recomputes its id in a second traversal (harmless), and the new Call/Proposal HASH_V0 goldens are pinned only via the OCaml path — the native `code.hash` shares the serialization by construction but isn't exercised against these specific identities; a one-line parity pin would close that.

jbwinters:
Review follow-up

The two-tier validator split is intentional. PR #32 constructs no proposal, so the artifacts-aware check does not apply there; the live gate must call `validate-proposal-artifacts` before consent or action. The optional dedicated native identity vector remains a hardening follow-up because the shared canonical serialization is already parity-tested.

PR #24, opened 2026-07-15
SC.7: add cooperative cancellation

jbwinters:
## Why this PR exists

A task can be cancelled, but cancellation must arrive at predictable safe points. Delivering it at an arbitrary instruction could interrupt an update halfway through. Never delivering it could let cancelled work continue forever.

## What changed

- Delivered cancellation when a task waits, yields, or pauses for a routed effect.
- Delivered immediately when the target is already suspended.
- Made cancellation terminal before any later user action can run.
- Transferred each suspended one-use continuation to the destruction callback exactly once.
- Returned awakened waiter handles in stable registration order.
- Preserved scope cleanup after normal cancellation, repeated delivery, and callback failures.
- Defined the callback-failure contract: the exception propagates, but the task stays Cancelled and the scheduler never re-owns or re-drops the continuation.

## Why it was done this way

Cooperative cancellation gives the runtime clear boundaries where program state is consistent. It remains deterministic and avoids interrupting evaluator work in the middle of an operation.

The result type does not contain a continuation for a cancelled task. That makes “drop, do not resume to unwind” an ownership rule the type system enforces instead of a convention.

## How it was checked

- A 200-case property covers cancellation delivery and ownership.
- Tests cover wait, yield, routed-effect, already-suspended, self-cancel, pre-cancel, repeated delivery, waiter order, and raising destruction callbacks.
- Native sanitizer stress covers teardown of 4,096 continuation-shaped values.
- Build, formatting, generated documentation, all 13 SC.7 overlay hashes, and the strict overlay checker passed.

## Review and merge notes

This PR is based directly on `main`; PR #21 is merged. It adds policy-independent cooperative cancellation, not the runnable queue or final Async root handler.


jbwinters (commented):

Built the branch: 573 tests green (120s), including the new cancellation suite (7 cases + 200-case QCheck property) and the ASan cancelled-continuation teardown probe. This resolves the drop-vs-unwind question I raised on the SC.5 PR, and resolves it well:

- v0 is drop-don't-unwind, and it's type-enforced rather than conventional: every cancelled outcome variant (`Boundary_cancelled`, `Await_cancelled`, `Yield_cancelled`, `Effect_cancelled`, `Cancel_caller_cancelled`) carries no resume field, so a caller cannot resume a cancelled continuation even on purpose. All three delivery paths funnel owned tokens to the caller-supplied `drop`. docs/concurrency.md states the consequence plainly — no user expression after delivered cancellation runs, dropping releases memory not external resources, bracket/with-file is the required C1 idiom — and `test_bracket_cleanup_without_post_cancel_step` demonstrates exactly that split.
- Delivery is confined to the three frozen points: `request_cancel` only sets a flag; terminalizing delivery happens at `at_cancellation_point` (await/yield/routed-effect) or `deliver_cancel` on an already-suspended target. A running task is never interrupted between boundaries, and a pending cancel preempts a routed effect's action before it runs (tested).
- Idempotency, self-cancel (well-defined via the caller's own routed-effect boundary), and once-token discipline (token lives in exactly one place; terminalize clears entry.resume; no double-consume path) all verify.

Findings, none blocking:

1. The one SC.0 obligation not directly pinned: no test cancels a target that has a registered awaiter and confirms the awaiter wakes and observes `Cancelled` — every cancellation test asserts an empty awakened list (test/test_cancellation.ml:24,38,113,140,162,180). The wake path is validated for completion/failure, so this is composition coverage, but it's the observability clause of the freeze and worth one test.
2. scheduler_core.ml:314's checked-out-already-cancelled branch is currently unreachable through the public API and untested — fine as forward defense, just unverified.
3. `cancel` checks the caller's own boundary twice for a runnable non-self target (structured_scope.ml:137,158) — intentional and harmless, but a one-line comment would save the next reader the trace.

Also noting for the record: single-task cancel doesn't cascade to the target's subtree — propagation is only via scope close, and the docs list scope failure policy as future work. Expected for v0.

jbwinters:
Review follow-up

Addressed in `ddea01c`:

- Added a target-with-registered-waiter cancellation test.
- Pinned waiter order, lifecycle, immutable Cancelled observation, and stale-boundary behavior.
- Documented why the caller boundary is checked again for runnable targets.

The later PR #30 also exercises the awakened-waiter path through the real global scheduler.

PR #25, opened 2026-07-15
GM.3: add validated governance policies

jbwinters:
## Why this PR exists

Governance needs a clear rule for deciding whether a call may run live, may run only as a simulation, or must be refused. Those rules include risk thresholds and confidence values, so invalid numbers or ambiguous boundaries could lead to unsafe decisions.

## What changed

- Added validated `LivePolicy`, `DryPolicy`, `StoredPolicy`, and `BoundPolicy` values.
- Confidence must be finite, and policy thresholds must be in a valid order.
- Every valid combination produces one total, predictable verdict.
- Policies have exact stable hashes.
- Forged or invalid policy bindings are refused.
- Tests cover full finite grids, numeric edge cases, and documentation examples.

## Why this approach

Policies are ordinary checked values instead of hidden host settings. That makes a decision reviewable, hashable, and reproducible.

Validation happens when a policy is built, so later gate code can rely on the policy invariants instead of repeatedly checking them.

## How it was checked

- 614 compiled tests.
- Focused governance, approval, taxonomy, ring, and prelude suites.
- Numeric-boundary and forged-binding properties.
- Full build, formatting, documentation, tier, and exact-manifest checks.
- Independent final review.

## Review and merge notes

This PR is stacked on PR #23. It also carries an already-validated concurrency test correction from its updated dependency base.

Review the governance stack through PR #23 first. Do not merge this PR until the owner gives direct approval.


jbwinters (commented):

Built the branch; governance-core suite green (14 tests, including the full 4×4 threshold grid — 160 live verdicts, 24 dry, forged-binding rejections, exact HASH_V0 goldens, and the numeric-boundary property), tiers cram green, GM3 manifest verifies. The full forced suite exceeded my harness timeout, so evidence here is the targeted suites covering everything this PR touches plus fmt/build — for a prelude-only change that's the right coverage.

The charter §6 properties all verify in code, not just prose:

- Forbidden is matched before any confidence/threshold logic (21-governance-core.jqd:555), so it absorbs even under `auto-up-to = Forbidden`.
- The under-confidence branch returns only Ask/Block — never Allow — and thresholds are inclusive with `auto-up-to <= ask-up-to` enforced at construction.
- Dry verdicts are Block/Simulate/NoSimulation only; no Allow, no live fallback.
- Purity is objectively pinned: all 11 new terms land in the pure tier (193→204), so the laws provably carry no effects.
- Nothing trusts a carried hash: `bind-stored-policy` recomputes, `validate-bound-stored-policy` re-derives and compares, and both verdict entry points front the validators, so a forged or invalid policy can be represented (public kernel constructors, per the documented backstop) but fails closed to InvalidDecision before reaching a verdict.

One scope note to carry consciously: there's no smart bridge from a validated StoredPolicy to a verdict — execution runs only through the direct BoundPolicy paths, and a caller must pattern-match `stored-live`/`stored-dry-run` and re-bind manually. The docs say the gate consumes these verdicts, so this looks like deliberate sequencing for GM.6/GM.7; confirm the gate PRs actually close that loop rather than each caller hand-rolling the match.

jbwinters:
Review follow-up

The StoredPolicy-to-execution bridge is closed by PR #32: the dry gate validates and consumes `governance.dry-policy-verdict` directly, including the missing-simulator refusal. The corresponding live-policy bridge remains explicitly owned by the live gate task rather than hand-rolled by callers.

PR #26, opened 2026-07-15
GM.5: add deterministic Judge handlers

jbwinters:
## Why this PR exists

A governed call needs an assessment before policy can decide what to do. The assessment may come from fixed rules, a test script, or an explicit model-backed handler, but all of those paths must produce the same checked value and behave predictably when input is missing or malformed.

## What changed

- Added the one-use `Judge.assess` operation.
- Added rule-based, fixed-result, scripted, and explicit inference-model handlers.
- Repeated runs with the same inputs produce the same assessment.
- Script exhaustion and malformed model output fail closed.
- The model path must be chosen explicitly; there is no hidden inference call.
- Effect rows, taxonomy metadata, and the Judge interface hash are pinned.

## Why this approach

Judge supplies an assessment; it does not make the final live/dry/refuse decision. Keeping assessment separate from policy makes both parts easier to review and test.

Multiple handlers support production, tests, and offline evidence without changing the public operation.

## How it was checked

- 619 compiled tests.
- Complete focused Judge, prelude, and taxonomy suites.
- Tier and native-effect command-line transcripts.
- Hostile one-use native coverage for all 16 relevant cases.
- Full build, formatting, documentation, and exact-manifest checks.

## Review and merge notes

This PR is stacked on PR #22. It stays separate from the policy PR so the later gate can combine two independently reviewed pieces.

Do not merge this PR until the owner gives direct approval.


jbwinters (commented):

Built the branch: 619 tests green, judge suite 9/9, both changed crams pass, and the GM5 manifest verifies (21 files OK). The four properties that matter check out:

- `judge.assess` is `once` end-to-end: declared in the op, listed in operation-modes.manifest, asserted as `Kernel.Once` in tests, and the once-hostile generator now includes it (16/16 rejecting identically with E0816 on both engines).
- Determinism is typed, not promised: `judge.rules` types its rule parameter with an empty effect row, so a rule provably can't read clock/RNG/env; fixed/scripted consume fixed inputs; byte-identical replay is asserted. The one nondeterministic handler (`judge.model`) keeps `Infer` in its outward row, so the nondeterminism is visible in the signature.
- Charter separation holds: outward rows are `{Throw | e}` / `{Infer, Throw | e}`, and a Net-using rule is a type error — authority can't be laundered behind Judge.
- Identity bookkeeping is coherent: the interface hash is pinned identically in all five locations (TSV, taxonomy doc, effect_registry, golden, recomputed-from-store test), and the implemented/reserved counts move 14→15 / 11→10 consistently.

One structural note to carry forward rather than fix here: the Assessment carrier has no call-id or policy-id (matching the charter schema exactly), so at this layer an assessment is transplantable between calls by construction — `judge.fixed` does it deliberately. The binding that prevents transplant in anger lives in the gate's Evaluated entry and the proposal hashes, which are sibling deliverables. I'll hold the GM.6 gate PR to that explicitly.

Prose nit: the evidence doc and stdlib section lean on negation-as-emphasis ("it does not introduce parallel...vocabulary", "not posterior judgment") several times; each instance is factual, but the pattern recurs enough to be worth one softening pass.

jbwinters:
Review follow-up

Addressed in `201965d`: softened the repeated negation-as-emphasis wording in the evidence, prelude comment, and stdlib documentation. Assessment remains call-independent by charter; PR #32 binds the freshly computed assessment to canonical call and policy IDs in the pre-action `Evaluated` audit entry.

PR #27, opened 2026-07-15
SC.8: add deterministic scope policies

jbwinters:
## Why this PR exists

When several child tasks run in one scope, the scope needs a clear rule for failures. Some programs should stop after the first failure. Others should finish every child and collect all results. Both choices must behave the same way every time.

## What changed

- Added typed `fail-fast` and `collect` policies for task scopes.
- Made fail-fast remember the first non-success result in deterministic scheduler order.
- Cancelled unfinished siblings in stable input order.
- Made collect mode preserve the caller’s original input order.
- Buffered waiter handles awakened by cancellation and exposed a drain-once handoff to the scheduler.
- Preserved awakened waiters even when the same target’s destruction callback throws.
- Continued every sibling cleanup, then re-raised the first callback exception with its original backtrace.
- Added pinned diagnostics and generated determinism properties.

## Why it was done this way

Failure handling is an explicit policy instead of an accident of queue timing. Code can choose the behavior it needs, and a reviewer can see that choice in the source and types.

Awakened waiters need a separate one-time handoff because making a waiter runnable without returning it to the global queue would silently strand work.

## How it was checked

- 37 command-line transcript files and 25 documentation examples across 8 documents passed.
- Two 200-case policy properties and repeated deterministic traces passed.
- Tests cover first-failure order, collect order, sibling cancellation, waiter registration order, drain-once behavior, repeated delivery, and raising destruction callbacks.
- The exact same-target raising-drop regression proves the awakened waiter stays runnable, owns its resume, and is handed off exactly once.
- Build, formatting, generated documentation, all 15 SC.8 overlay hashes, and the strict overlay checker passed.
- A fresh independent review of the corrected exact head found no blockers.

## Review and merge notes

This PR remains stacked on PR #24 until cooperative cancellation lands. It adds deterministic scope failure policies, not the final runnable queue or Async root handler.


jbwinters (commented):

Built the branch: 573 tests green (123s) including the new scope-policy suite and cram (byte-identical repeat run), manifest verifies against the updated base pointer. The four contract questions all check out:

- First failure is scheduler decision order, matching the frozen `first_failure`, and the test proves it against the tempting wrong answer: a later-spawned child recorded at decision 20 beats the earlier-spawned one at 21. Strictly-increasing decision numbers eliminate sequence ties, which is what makes the incremental selection equivalent to the contract's `(sequence, task_id)` ordering.
- Fail-fast cancels through the SC.7 model exactly: request for runnable siblings (observable at their next boundary), immediate delivery for suspended ones at the correctly mapped point, one `drop` per affine token, no double-drop, observation committed before cancellation and not rolled back if drop raises.
- Collect preserves input order including Cancelled entries, and leaves siblings uncancelled.
- The controller never resumes or schedules — policy choice can't change child semantics, only aggregation and cancellation.

Three findings to settle at SC.9 integration if not here:

1. `scope_policy.ml:74-78` discards the awakened-handle list `deliver_cancel` returns (`let* _ = ...`), unlike every other SC.7 boundary which surfaces it. If a fail-fast-cancelled suspended sibling has waiters, they become runnable inside scheduler_core but are never reported to the driver — SC.9 must re-scan task_views to recover them. And the path is untested: no scope-policy test creates an await edge among siblings, so the awakening branch never fires. Either surface the handles or pin the re-derivation contract with a test.
2. First-non-success selection is re-derived incrementally rather than delegating to the frozen `first_failure`, and the 200-case property covers Collect only — fail-fast's selection rests on one hand-crafted case. A property asserting agreement with `first_failure` over random multi-failure scopes would close the single-source-of-truth gap.
3. Inherent cooperative-cancellation caveat worth one explicit line in the SC.8 prose: a child that never reaches a suspension point can't be preempted, so fail-fast can't guarantee scope drain against a spinning child.

jbwinters:
Review follow-up

Addressed in `56311b7` and PR #30:

- Added a 200-case property proving incremental fail-fast selection agrees with frozen `first_failure`.
- Documented that cooperative cancellation cannot preempt a child that never reaches a boundary.
- PR #30 now has a real integration test where fail-fast cancellation awakens a registered waiter, returns it to the global queue, and drains it without later world work.

PR #28, opened 2026-07-15
GM.4: add exhaustive governance policy laws

jbwinters:
## Why this PR exists

The governance policies added in PR #25 use risk levels, confidence values, and safety thresholds to decide whether an action is allowed, needs approval, or must be blocked. A few hand-written examples are not enough to show that every valid combination follows those rules.

## What changed

- Added eight separately named laws for policy decisions and policy identity.
- Tested every valid threshold and risk combination with representative confidence values.
- Added checks for exact numeric boundary cases.
- Added small counterexamples that explain what went wrong when a law is broken.
- Tested both sampled and exhaustive runs.
- Verified that cold-cache and warm-cache runs return the same answer.
- Added a deliberately broken policy witness to prove the test harness catches real mistakes.

## Why it was done this way

This PR adds evidence without changing runtime behavior. Keeping these laws separate from the policy implementation makes the review easier: PR #25 defines the rules, and this PR tries to break them from many angles.

## How it was checked

- Full local suite: 655 compiled tests passed.
- Exhaustive supports up to 4,000 generated cases.
- Cold-cache, cache-hit, and source-immutability checks passed.
- Build, formatting, documentation, command-line tests, and evidence hashes passed.
- Independent review found no blocker.

## Review and merge notes

This PR is now based directly on `main`. It changes policy tests and release evidence; it does not add handlers or world access.


jbwinters (commented):

Built the branch: full suite 614 green (145s), both Warp lanes pass (sampled seed 145 and exhaustive with counts reproduced exactly: 4000/2000/1000/500/150/40/20/5 + 6 boundary checks), GM4 manifest verifies. Since this PR is entirely laws, I judged it on whether the laws are right, exhaustive, and non-vacuous:

- Right: the eight properties map one-to-one onto charter §15, the monotonicity laws check verdict rank against the correct ordering, sensitivity/stability are split correctly (each semantic field change flips the hash; summary changes and reformatting don't), and dry-totality proves dry never yields Allow/Ask.
- Exhaustive: the finite grids are fully enumerated under --exhaustive, and real-valued confidence is handled the right way — 5-point grid in the properties plus the two adjacent representable doubles around 0.5 in the unit case.
- Non-vacuous, proven rather than assumed: I mutated the GM.3 implementation six ways (Forbidden→allow, gte?→gt? at the boundary, inverted confidence gate, dropped arguments from the call subject, dry no-simulation→simulate, validator always-accepts) and the suite caught every one, with the boundary mutation caught only by the numeric-edge case — which validates the grid+edge design.

Two findings, first one worth a follow-up law:

1. The min-confidence axis is pinned at 0.5 in both `gm4-confidence-monotonicity` and `gm4-policy-tightening` (policy built with `(lit 2)`), and `is-tightening` compares only risk ranks — so tightening via a raised confidence threshold, which is a legitimate tightening, is never exercised, and thresholds 0.0/0.25/0.75/1.0 are never used as thresholds. The charter's "lowering thresholds never increases the allowed set" arguably includes this axis.
2. The whitespace half of `gm4-call-hash-stability` is near-tautological — the quote reader normalizes both forms to identical Code before hashing, so it exercises the reader rather than a distinct serialization path. The summary-change half carries the law.

Scope note: proposal/assessment identity laws are correctly absent — §15 names only Call ID stability/sensitivity — but since the task title says "identity," worth remembering that proposal-identity laws still have no property coverage anywhere.

jbwinters:
Review follow-up

Addressed in `1a5a8d5`: added a separate 500-case exhaustive property for tightening the minimum-confidence threshold, updated the sampled/exhaustive counts and cache expectations, and added mutation checks so the new axis is non-vacuous.

PR #29, opened 2026-07-15
Effect taxonomy: audit, secret, approval, and release freeze

jbwinters:
## Why this PR exists

PR #14 defined the trusted effect catalog, but several safety-sensitive effects were still only planned. Audit, Secret, and Approval need real values and handlers before governed membranes can rely on them. The finished catalog also needs release evidence showing that every layer agrees.

## What changed

- Added deterministic audit entries and hash-chain verification.
- Added opaque Secret values and explicit handlers.
- Redacted secret data in interpreter and native output paths.
- Added approval proposals whose identity is bound to the exact reviewed content.
- Added deterministic approval handlers.
- Completed and froze the taxonomy as 15 implemented effects plus 10 reserved slots.
- Added exact schemas, identities, handler rules, authority differences, and reviewer guidance.

## Why this approach

Audit records should reveal tampering, secret values should not become ordinary printable text, and approval must apply to the exact proposal that was reviewed. Building these together closes the effect-taxonomy epic with one consistent safety story.

Reserved entries remain identity-free until a future interface is actually approved.

## How it was checked

- 611 compiled tests and 34 command-line transcripts.
- Exact agreement between the TSV catalog, typed registry, and loaded prelude operations.
- Stale-hash, redaction, proposal-binding, and authority-difference tests.
- Native comparison: 69 identical programs, 8 expected refusals, 0 failures.
- Native leak checks, 1,000 generated native cases, installer/demos/CLI/docs, and exact release reproduction.

## Review and merge notes

This PR is the effect-taxonomy epic boundary and is stacked on PR #14.

Review PR #14 first. Do not merge this PR until the owner gives direct approval.


jbwinters (commented):

Built the branch and ran the suite: 611 tests green (113s), including the tamper/mutation QCheck suites, the 300-case redaction property, the approval replay/mismatch suites, and the native gauntlet. Reviewed the three effects as security mechanisms:

- Audit chain is genuinely tamper-evident: each digest covers a domain tag + the predecessor's raw 32 bytes + canonical entry bytes (src/audit_chain.ml:115-118); verification re-derives every digest from genesis rather than trusting stored hashes, and tail truncation is caught because `finish ~expected_head` commits the caller to the whole chain (E1305, tested). Distinct diagnostics for reorder/removal/alteration/duplication/version/malformed.
- Secret opacity is structural, not string filtering: swept every channel — `Value.show`, native `jq_show`, both `debug.inspect` paths, checker-level E0818 blocking Secret→Text, payload-free provider errors, and no code.hash path can touch secret bytes (VSecret is never VCode). Both engines render the fixed `<secret redacted>` marker.
- Approval binding: proposal-id is HASH_V0 over the canonical proposal-v1 encoding of the seven review fields, excluding the carried id; decision validation recomputes rather than trusts; each field change invalidates the binding (tested per-field) while metadata-only changes don't. Consent laws are enforced in handler code. Byte-identical replay is possible by design — single-use is correctly deferred to the membrane layer and documented.

Three notes, none blocking:

1. Secret bytes aren't zeroed on free — JQ_SECRET shares the no-op free path (runtime/jq_rc.c:69), so freed heap retains plaintext until overwritten. The PR never claims scrubbing (jq_value.h:56 scopes the guarantee to rendering), so this is a hardening opportunity, not a violated claim. Worth a line in the threat-model docs either way.
2. test/test_secret.ml:108 — the marker-present assertion has a `|| String.contains rendered '<'` fallback that passes on nearly anything. The paired leak-absence check on line 105 carries the real weight, but the first half is close to vacuous.
3. `approval.policy-auto` trusts the injected classifier's verdict — the verdict isn't bound into the proposal (only the assessment hash is), so a wrong classifier returning `allow` on an Ask proposal yields approved. Documented contract, but it's the trust boundary to watch during membrane integration (GM.5/GM.6).

jbwinters:
Review follow-up

Addressed in `42ffa86` and `24c8f60`:

- Documented that Secret carriers are redaction-safe but not memory-scrubbed on free.
- Documented the trusted-classifier boundary and the live membrane's validation obligation.
- Replaced the near-vacuous redaction assertion with an exact marker check.
- Made the concurrent truncation test assert the actual fail-closed contract without requiring a sampling race to produce one specific diagnostic.

PR #30, opened 2026-07-15
SC.9: deterministic round-robin Async scheduler

jbwinters:
## Context

The earlier structured-concurrency changes define tasks, scopes, waiting, cancellation, and failure policies. They still need a real scheduler that decides which ready task runs next.

Without one shared rule, the same program can behave differently depending on where it is evaluated. That makes failures hard to reproduce and scheduler limits hard to explain.

## What changed

- Added a FIFO round-robin scheduler. Ready tasks take turns in the order they became ready.
- Made nested scopes share one scheduler, one decision order, and one set of limits.
- Made waiting tasks return to the queue in a stable order, including waiters awakened while cancellation cleanup is running.
- Added a real deadlock result when work remains but no task can run.
- Protected routed effects and suspended work from hostile mutation.
- Made each evaluation start with fresh private task ownership, even when callers reuse an evaluation context.
- Added a narrow default grant for only Jacquard's exact built-in `Async` definition.
- Added tests for yielding, waiting, cancellation, failures, nested scopes, escaping tasks, deadlocks, bounds, and hostile checker or Warp paths.

## Why it was done this way

Round-robin scheduling is small, predictable, and easy to reproduce. A single scheduler shared by nested scopes prevents each scope from inventing its own clock or queue.

The automatic `Async` grant is intentionally narrow. The normal interpreted CLI, prelude evaluation, and Warp use this scheduler and recognize only the frozen built-in `Async` hash. Raw `Eval.run_expr` is still unscheduled, and the native runtime does not secretly add a root Async scheduler.

Task constructors and scheduler ownership stay private, so user programs cannot forge handles or directly control the queue.

## How it was checked

- 674 compiled tests.
- 38 command-line transcripts.
- 25 documentation examples across 8 documents.
- Full build, test, formatting, documentation, runtime, and native checks.
- All 16 active release manifests and the structured-concurrency overlay check.
- Independent review of the exact final commit, including exception identity, backtrace preservation, and ownership restoration.

## Merge note

PR #27, which adds the deterministic scope policies this scheduler uses, is now merged. This PR is the next change in the structured-concurrency stack.


jbwinters (commented):

Built the branch: 578 tests green (162s), including the 128-run byte-stable golden traces. This is the integration milestone and it holds together:

- Queue discipline matches the frozen contracts, verified line-by-line against the golden trace: spawn appends child-then-parent, woken awaiters append at completion, yield requeues to tail. Determinism is structural — no clock/RNG/hash iteration anywhere in the scheduler path, cache keyed on canonical program hash.
- The threads carried from earlier reviews mostly close: the dead SC.3 `scheduler_task_value` seam is deleted and replaced by a live capability-gated mint (`Scheduler_core.task_value`, called on every spawn, with `public-api.t` proving the capability modules are installed-private) — real programs now spawn/await/cancel real Task values end to end, which also retires the SC.6 caveat about Obj-forged escape tests. Awaiter-observes-Cancelled is now pinned in both the OCaml and Warp lanes.
- The hostile suite earns the name in the OCaml lane: stale-Task reuse across runs, a hostile root handler mutating a suspended Once resume to smuggle a foreign Task, escape refusal, cancel-preempts-routed-effect with empty output buffer.

Correction for the record: I initially suspected the SC.5-SC.8 unit suites weren't executing before this PR (no test_jacquard.ml registration). That's wrong — they ran via the `run ()` calls in test_effect_taxonomy.ml:773-776 all along; this PR just promotes them to first-class registered suites. No dormant-coverage issue.

Findings, none blocking:

1. The frozen contract relations are decorative relative to the running code: `requeue_after_*` has no production call site (the queue is reimplemented inline), `decide_round_robin` only renders the trace while the executed head is `List.hd runnable` with no assertion tying them, and `first_failure`/`order_completions` are still consumed only by tests. Behavior provably matches today (goldens), but nothing structurally prevents drift between the frozen relations and the inline implementations. One conformance property per pair would make the freeze load-bearing.
2. The driver's `Await_deadlocked` branch (round_robin.ml:392-396) is unreachable from surface Jacquard and untested at integration level — correctness rides on `fail_cycle` having terminalized already, which it does.
3. Deliberate-decision check: `granted_hashes` now auto-grants the frozen Async effect unconditionally — Async no longer needs `--allow`. Defensible (the scheduler is Async's handler and child world-effects still need their own grants), but it's a capability-surface change that deserves a sentence in the docs/evidence rather than being implicit.

jbwinters:
Review follow-up

Addressed in `8d171ab`:

- Production scheduling now selects the task named by `decide_round_robin` and checks spawn/suspend queue updates against the frozen `requeue_after_*` relations.
- Added real evaluator coverage for the `Await_deadlocked` branch.
- Added fail-fast coverage proving an awakened waiter is returned to the global queue and drained.
- Documented the narrow automatic Async grant; child Console/Fs/Net effects still require explicit grants.
- Added a checkout bracket so an unsettled affine token returns to scope ownership on normal return, diagnostic error, or host exception.

The full branch gate passes 578 tests plus docs, formatting, and the structured-concurrency manifest.

PR #31, opened 2026-07-15
GM.9: define typed Workspace governance facade

jbwinters:
## Why this PR exists

Governed membranes need a concrete example of how real tools should be described before they are allowed to touch the outside world. File reads, file writes, and network requests are useful examples because a reviewer needs to see the exact path or URL, the authority being requested, any safe secret references, and a summary of the result.

## What changed

- Added a typed `Workspace` interface for reading files, writing files, and fetching a URL.
- Made each helper build a reviewable governed call for one exact operation.
- Kept paths and URLs readable so a person can review them.
- Replaced file contents and request bodies with stable hashes in summaries, so sensitive or large data is not copied into review records.
- Represented secret metadata with `SecretRef`; these helpers never accept a raw secret value.
- Added separate safe result summaries for reads, writes, and fetches.
- Corrected canonical authority ordering so both the interpreter and native runtime use the frozen effect-taxonomy order instead of sorting rendered text.
- Reconciled the facade with all current governance changes and release evidence.

## Why it was done this way

Typed helpers make the requested action clear. A reviewer can tell whether code wants to read a file, write a file, or use the network instead of seeing one vague string-based tool call.

The helpers stop at building and summarizing governed calls. They do not perform live I/O. Later dry-run and live handlers can therefore share this reviewable identity boundary without being hidden inside the facade.

## How it was checked

- 668 compiled and property tests, 36 command-line transcript files, and 25 documentation examples across 8 documents passed.
- Tests prove that meaningful arguments change a call ID while presentation-only wording does not.
- Secret-redaction, forged-spec, unknown-effect, and exact taxonomy-order cases passed.
- Interpreter and native authority comparisons agree across all 26 catalog positions and Resource ordering edge cases.
- The complete native walk reports 69 identical programs, 8 expected refusals, and no failures.
- Build, formatting, generated documentation, all 173 GM.9 overlay hashes, and the complete-overlay checker passed.

## Review and merge notes

This PR is based directly on `main`; its required governance gate is now merged.

Important: this facade PR also contains the necessary base-level correction to canonical authority ordering. Ordering now follows structured taxonomy fields instead of punctuation-heavy rendered text, keeping interpreter and native checks consistent.


jbwinters (commented):

Built the branch: 632 tests green (151s), including the new workspace suite, the g38 native differential twin, and 21/21 once-hostile cases. The four charter axes all verify:

- Typed per-operation, with the anti-pattern actively blocked: three `once` operations over a closed sum with typed argument shapes, no universal string-call operation (`test_no_generic_call_bypass`), and a forged operation spec can't type into the normalizers because they take typed args and recompute their own spec.
- Identity is entirely GM.2's: normalizers route through `governance.make-call`, which Store-resolves the operation-id and hashes the canonical subject — the facade computes no identity itself, and the presentation summary stays identity-inert.
- Summarizers are total, deterministic, and secret-safe structurally: no input type mentions Secret, payloads enter argument Code only as digests, and `tool-error-show` reduces hostile driver text to a fixed closed label (tested by injecting a secret into driver-failed).
- The boundary is honest: no driver, no simulator, all public schemes closed-pure, unhandled Workspace is E0814 with a pointed message, and `--allow workspace` is refused with E0703 — the facade can only ever be satisfied by an in-program membrane.

Four notes, none blocking:

1. Be aware this PR bundles a base-level semantic change: canonical authority ordering moves from rendered-Code lexical to frozen taxonomy position (`governance.authority-after?`). It's necessary — under the old order the fetch envelope [Net, Secret] would sort wrong and be rejected — and the OCaml/native catalogs were cross-checked to agree on all 26 rows, but it changes GM.1/GM.2 validation semantics from inside a GM.9 PR, so it deserves its own line in the merge notes.
2. The three blessed interface hashes are hardcoded string literals in the prelude (23-workspace.jqd:45,52,59) — a fourth copy alongside TSV/OCaml/C. A typo fails closed through validate-authority plus goldens, and there's no term-level way to derive them, so acceptable — but the copy count keeps growing.
3. Those same three sites match only `ok` from `hash.parse` with no `err` clause — safe with valid literals, but a mistyped one would die as a match failure instead of a clear diagnostic. Only occurrence of that idiom in the prelude.
4. The `net fetch → fetch` short-name alias in src/prelude.ml is tangential to the facade; tests cover it, just scope creep.

jbwinters:
Review follow-up

The PR description now carries an explicit merge note: canonical authority ordering changes from rendered-Code lexical order to frozen taxonomy order, which is required for the typed Workspace fetch envelope. The hardcoded frozen hashes, fail-closed parse shape, and short-name alias are accepted release seams; their goldens and validators catch drift.

PR #33, opened 2026-07-15
SC.10: add versioned scheduler traces and strict replay

jbwinters:
## Context

A concurrency bug may appear only when tasks take turns in one particular order. Developers need to record that order and replay it exactly. If the program or trace changes, replay must stop safely instead of silently running a different schedule and touching the outside world.

## What changed

- Added a canonical version-1 schedule-trace format.
- Recorded the program hash, scheduler, policy, task and decision limits, task creation, exact runnable queue, chosen task, and operation at every step.
- Added CLI options to record a run, replay it strictly, or fork it at one decision.
- Made replay check task creation before allocating a task and check routed operations before calling outside-effect handlers.
- Made missing, extra, reordered, impossible, damaged, or noncanonical events fail with E0908.
- Refused old and unknown trace versions.
- Bounded trace headers, line sizes, total bytes, and event counts while loading.
- Added exact errors for missing, unreadable, oversized, and unwritable trace paths, including failures that happen while the trace is being closed.
- Included the trace format version and run limits in scheduler cache identity.

## Why it was done this way

Replay is useful only when drift cannot fall back to normal scheduling. The controller therefore compares the whole runnable queue, selected task, and operation at every step.

Checking before allocation or outside-effect callbacks matters. When a trace no longer matches, Jacquard stops before the changed path can read a file, use the network, print routed output, or perform another outside action.

The format is versioned and canonical so a valid program and trace replay the same way across machines.

## How it was checked

- 677 compiled tests.
- 39 command-line transcripts.
- 25 documentation examples across 8 documents.
- Golden record/replay bytes and byte-identical fork replay.
- Refusal tests for missing, extra, reordered, impossible, old-version, unknown-version, and noncanonical traces.
- Bounds tests for tasks, decisions, queues, lines, total input, and event counts.
- Tests proving routed-operation drift is caught before callbacks or output.
- Missing, unreadable, delayed-write, and close-time write-failure tests.
- Full build, formatting, documentation, runtime, native, and all 16 active release-manifest checks.

## Merge note

PR #30, which adds the deterministic round-robin scheduler used by recording and replay, is now merged. This PR is the next change in the structured-concurrency stack.


jbwinters (commented):

Built the branch: 581 tests green (150s), including the new schedule-trace codec suite, the record/replay/drift/fork round-robin cases, and the replay cram. The properties that make record/replay trustworthy all verify:

- Versioning is a hard gate: any format other than 1 is refused, unversioned input is refused, no migration or guessing, and the format version is folded into the cache key.
- The trace pins the schedule and nothing else: program HASH_V0, scheduler id, policy, bounds, structural task IDs, and each decision's exact ordered queue + chosen task + operation. No wall-clock, no paths, no routed arguments. Cross-run byte-stability is asserted with `cmp` in the cram.
- Strict replay is total, with no fallback path: every creation is checked before allocation, every decision's full ordered queue is compared, every operation is observed before any world callback fires (the drift tests pin `calls = 0`), and EOF/leftover both error. A changed program dies at the header hash.
- Canonical encoding is enforced by re-serialize-and-byte-compare on parse, which also catches the `int_of_string` lenient spellings (0x, underscores, leading zeros). Truncated/malformed input fails closed with sensible size bounds.

One nit: `write_schedule` in bin/main.ml wraps the write in `Fun.protect ~finally:close_out_noerr`, and since real traces are smaller than the channel buffer, the actual flush happens inside `close_out_noerr` — an ENOSPC or I/O error at flush is swallowed and the CLI exits 0 having written a truncated file. The partial file fails closed on any later replay, so no incorrect replay is possible; the only harm is a false "record succeeded". An explicit `flush`/`close_out` inside the protected body would surface it.

jbwinters:
Review follow-up

Addressed in `7574ea8`: schedule recording now performs an explicit checked `close_out` inside the protected body, so buffered flush and close errors reach the CLI. `close_out_noerr` remains cleanup only. A `/dev/full` cram regression proves the command no longer reports false success.

jbwinters:
Restack note

After PR #30 merged, this branch was attached to merged `main` with ancestry-only commit `009bcf74ceb42c15b5b5d1a31dec6dc4d622da37`.

The tree hash remains exactly `a77d5255a0ce354ed5738c0b0531c4ff103abc12`, identical to independently reviewed head `b5029ad03fc8acb761161296747c44663be1ab4f`. No source, test, documentation, or generated file changed during the restack. Fresh exact-head CI is running because branch protection requires the new `main` commit in this PR's ancestry.

PR #35, opened 2026-07-17
SC.13: freeze scoped channel semantics

jbwinters:
## Context

Structured tasks need an exact channel contract before the scheduler can safely implement channels. Without one shared contract, two implementations could disagree about buffering, close behavior, cancellation, wake order, or who owns a channel handle.

This PR freezes those rules. It does not add the channel runtime yet.

## What changed

- Defined the four once-only Channel operations and pinned their exact HASH_V0 identities.
- Defined capacity rules for rendezvous and buffered channels.
- Defined FIFO send, receive, wake-up, and fan-in order.
- Defined close behavior, including buffered-value draining, blocked waiter results, and repeated close.
- Defined how cancellation removes blocked senders and receivers without losing or duplicating values.
- Made ChannelHandle ownership local to one scheduler run and structured scope.
- Required matching structured-scheduler admission; `--allow channel` cannot grant Channel operations.
- Added readable rendezvous and buffered traces.
- Added positive and negative checker fixtures, documentation examples, and release evidence.

## Why it was done this way

Channels combine several rules that are easy to get subtly wrong. Freezing the behavior first gives the later runtime work one reviewable source of truth.

Keeping this PR contract-only also separates semantic decisions from the larger SC.14 implementation. The runtime implementer should not need to guess about ordering, blocking, close, cancellation, or ownership.

## How it was checked

- 687 compiled tests.
- 39 command-line transcripts.
- 27 documentation examples across 8 documents.
- Focused effect-taxonomy, surface-law, scheduler, replay, seeded, and exhaustive-schedule checks.
- Positive and negative channel type/effect fixtures.
- Exact rendezvous and buffered trace examples.
- Full build, test, formatting, documentation, runtime, native, and all 16 active release-manifest checks.
- Independent review of the final restacked head.

## Stack note

This PR is stacked on PR #39. It freezes SC.13 channel semantics; SC.14 remains the later runtime implementation task.

PR #41, opened 2026-07-17
SC.14: implement scoped typed channels

jbwinters:
## Context

PR #35 freezes the rules for typed channels, but it deliberately does not implement them. Without this change, Jacquard programs can describe the Channel effect but cannot open a channel, send a value, receive a value, or close the channel through the real scheduler.

SC.14 adds that runtime while keeping channels inside structured task scopes. A channel handle belongs to one scheduler run and one exact scope; it cannot be reused later or smuggled into another scope.

## What changed

- Added the exact Channel effect operations and prelude surface for `open`, `send`, `recv`, and `close`.
- Implemented zero-capacity rendezvous channels and positive-capacity FIFO buffers.
- Kept pending senders and receivers in deterministic FIFO order.
- Made close idempotent, preserved already-accepted buffered values, rejected blocked senders, drained the buffer, and then rejected remaining receivers.
- Stored suspended task IDs in channel queues while keeping every once-only continuation owned by the scheduler.
- Added prepare/commit boundaries so argument mapping, caller validation, suspension, and every wake are checked before channel or scheduler state changes.
- Made cancellation remove a channel waiter only after a real pending cancellation is proved, preventing suspended tasks from becoming orphaned.
- Required exact run identity, exact scope identity, and a live channel-registry entry for every operation.
- Rejected stale, forged, cross-run, cross-scope, and recursively escaped handles with the documented diagnostics.
- Integrated Channel decisions with deterministic traces, strict replay, cache identity, fail-fast and collect scope policies, and deadlock reporting.
- Kept native execution unsupported for Channel programs and updated the CLI refusal transcript instead of silently changing the native contract.
- Updated hashes, effect inventories, tutorials, evidence, release manifests, and exact transcript counts.

## Why it was done this way

Channel state and scheduler state must change as one operation. If a value were removed from a buffer before a waiter could be safely resumed, a failure could lose the value or strand a task. The prepare/commit design performs every fallible check first, then applies only non-failing mutations.

The scheduler remains the only owner of suspended continuations. Channel queues contain task identities and pending values, not a second copy of a continuation. That preserves Jacquard’s once-only ownership rules during wakeup, cancellation, failure, and scope cleanup.

Exact run, scope, and liveness checks make a channel handle useful only inside the structured scope that created it. This prevents hidden shared state from escaping its lifetime.

## How it was checked

- Full build and 700 compiled/property tests.
- 39 command-line transcript files, including updated store and native-effect refusals.
- 27 documentation examples across 8 documents.
- The two frozen SC.13 normative traces replayed field for field.
- Rendezvous and buffered FIFO tests, sender promotion, close/drain/idempotence, cancelled middle waiters, nested scopes, fail-fast, collect, and deadlock cases.
- Adversarial tests for raising value mappers, invalid caller lifecycle, wrong suspension identity, failure during later close wakeup, no-request and duplicate cancellation, raising cleanup, malformed IDs, and channel-ID exhaustion.
- A mixed-operation reference-model property with unique values and conservation/no-duplication checks.
- Escape tests for stale, forged, cross-run, cross-scope, and recursively nested handles.
- An allocation regression proving Task and Channel private-hash checks stay allocation-free across 100,000 hot-path calls.
- The 200,000-element sort witness and full native differential lane: 69 identical programs, 8 expected refusals, and no failures under the unchanged timeout.
- Zero-live-task, zero-resume, and zero-channel cleanup checks.
- Formatting, generated documentation, structured-concurrency manifest hashes, release evidence assertions, and whitespace checks.

## Stack note

This PR is stacked on PR #35, which freezes the typed-channel contract implemented here. It should be reviewed after #35 and will be attached to merged `main` only after the earlier scheduler stack lands.


jbwinters (commented):
**Sign-off: LGTM.** (Posted as a comment because GitHub does not allow the PR author to approve their own PR.)

Two independent reviews approve final head `022f0d72d9d304a0ba6189e4affdc8005b00adc7` against exact SC.13 base `1bf25e9bf27a86931eee91616acd19bf76b32f80` with no blockers or nits.

The reviews checked the atomic prepare/commit boundary, cancellation ownership, channel-open error domains, exact run/scope/liveness checks, FIFO rendezvous and buffering, close/drain/idempotence, fail-fast/collect/deadlock behavior, strict replay, seeded and exhaustive parity, cache behavior, native refusal, escape prevention, and zero-ownership cleanup.

Review findings were fixed before this PR opened: release evidence now reports the exact 700/39/27 inventories and includes the 12 channel-contract cases; stale “runtime deferred” wording is removed; seeded/replay/cache tests run a real rendezvous program; exhaustive mode proves four distinct worlds and the exact `{open, send, recv}` operation multiset; non-Channel effects remain refused before callbacks.

Independent validation passed the full 700-test suite, native/runtime lanes, 39 command-line transcripts, 27 documentation examples across 8 documents, formatting, generated docs, documented whitespace checks, and all 57 structured-concurrency manifest entries.

jbwinters:
**Sign-off: LGTM on the performance-fix head.** (Posted as a comment because GitHub does not allow the PR author to approve their own PR.)

The final follow-up changes this PR from `022f0d72d9d304a0ba6189e4affdc8005b00adc7` to `e1ffd22ec0289d40f334f9ebc1171cfb760759d0`.

The first CI run exposed a real interpreter slowdown in the 200,000-element sort witness. Task and Channel handle checks converted a SHA-256 hash to hexadecimal on every constructor check. That produced millions of short-lived allocations after Channel added a second check to the same hot path.

The new commit decodes each frozen private hash once when its module loads, then compares hashes directly. Malformed built-in constants still fail as `Bug_`-prefixed internal invariant errors. The release manifest now includes both contract modules so the evidence archive reconstructs the exact performance-sensitive code.

The regression test calls both predicates 100,000 times and permits less than 4 KiB of allocation. An independent reviewer transplanted that test onto the old head, where it failed specifically on the Channel check; it passes repeatedly on the new head.

Independent validation on exact head `e1ffd22ec0289d40f334f9ebc1171cfb760759d0` found no semantic or contract drift and passed:

- the full 200,000-element interpreter sort under the unchanged 180-second timeout, with output `200000`;
- the full native differential lane: 69 identical programs, 8 expected refusals, and 0 failures;
- the focused Channel, store, and effect-taxonomy suites;
- the full build, formatting, whitespace, manifest-coverage, and manifest-hash checks.

Measured full-sort time improved from 110.71 seconds on the old Channel head to 87.42 seconds locally. A separate independent run completed in 84.84 seconds. The pre-Channel SC.13 baseline was 95.65 seconds.

jbwinters:
## Exact-head sign-off

**Sign-off: APPROVE**

Reviewed exact head `4e91ca8164a9c3ae9cbb0d5ad1202c2682761010` against PR #35 head `df600233e7ba5ac5472928192b990995c7bb2662`.

The scoped channel runtime keeps FIFO order, close and cancellation behavior, exact run/scope ownership, and once-only continuation ownership. The allocation regression also passes for 100,000 private-hash checks.

A review initially found that 15 integration manifests did not list six new SC.14 files. That blocker was fixed by regenerating every manifest from its declared base. All 16 active manifests now have valid hashes and complete file coverage with no missing, extra, duplicate, or deleted paths.

Validation passed: full build, full test suite with 700 compiled tests, command-line and documentation lanes, generated docs, formatting, focused channel/scheduler tests, and the allocation checks.

jbwinters:
## Current exact-head confirmation

**Sign-off: APPROVE**

Confirmed PR #41 exact head `9cad4f050d50735a9c4415967312a7f2a055e886` against PR #35 head `282fc29140c180c231cdb3c64e163c6dc4adc366`.

The tree is byte-identical to the approved SC.14 tree. The current base is an ancestor, all 16 manifest hash and full path-coverage checks pass, and the worktree is clean. GitHub CI is still running.

jbwinters:
## Exact-head review after PR #35 merged

**Sign-off: APPROVE — no blockers or nits.**

This review covers the current PR head, `ac16cc1f51db99416bb8cedd641636b0077d8677`, based on the current `main` commit, `982a14a895757ff505e81e9301e14a553852f01c`.

The new commit only connects this branch to the latest `main` history. Its file tree is byte-for-byte identical to the previously approved SC.14 head, `9cad4f050d50735a9c4415967312a7f2a055e886`, so it does not change the scoped-channel implementation or any other repository content.

The fresh review also confirmed:

- Current `main` is an ancestor of this exact head.
- The ancestry-only commit has an empty file diff.
- All 16 active release manifests pass strict hashes and complete file-inventory checks.
- The structured-concurrency manifest checker passes.
- Eight focused structured-concurrency test groups pass, covering 28 scheduler, channel, scope, cancellation, trace, and exhaustive-schedule cases.
- Formatting passes.
- There are no unresolved review threads or outstanding findings.
- The exact-head worktree is clean.

GitHub CI is still running for this exact head. The PR should merge only after all three required checks succeed and GitHub reports a clean merge state.

PR #53, opened 2026-07-19
Add crash-safe single-use approval queue core

jbwinters:
## Context

The released governance gate binds a Decision to the exact Proposal it reviewed, but the shipped approval handler is deliberately scripted evidence. It does not make a reviewer decision globally fresh, persist consent across process restarts, or prevent the same returned Decision from being supplied again.

GM.13 needs a durable queue between an authenticated host workflow and the live gate. This PR lands the independently useful first half: a production-oriented OCaml storage core with one-time delivery. It does not install a `GovernanceApprovalV1` handler or claim that an outside-world action happens exactly once.

## What changed

- Added the public `Governance_approval_queue` host adapter for Submit, Decide, atomic Consume, inspection, and explicit recovery.
- Preserved the released GovernanceProposal and Decision Code identities. Allowed principals and the authenticated host actor are queue metadata; they do not change Proposal or Decision HASH_V0.
- Added a versioned content-addressed journal. Every transition stores a HASH_V0-linked canonical record line followed by a separate commit line; both phases are synced before success is acknowledged. Every successful existing-file non-Busy mutating result also establishes a file-and-parent-directory durability barrier, including exact retries and stale or pending observations; missing-file recovery returns the fixed empty genesis without I/O.
- Added fail-closed restart handling. Only a narrowly recognized physically uncommitted suffix after the last verified commit can be truncated; committed corruption is never auto-repaired.
- Made every Decision variant single-use. Consume records the exact Proposal and Decision IDs durably before returning Approved, Denied, or Escalate once; later delivery is Stale and cannot be reset.
- Combined a fork-aware process-local guard with a nonblocking whole-file lock, bounded regular-file reads, and pathname/inode checks. This covers competing OCaml Domains and processes without adding ambient file authority to Jacquard.
- Added stable E1520-E1526 diagnostics, thirteen adversarial compiled cases, GM.13A release evidence, and a reconstructible overlay manifest.
- Updated the successor release inventory from 742 to 755 compiled/property cases.

## Why it was done this way

The durable Consume-before-return order chooses at-most-once queue delivery. A crash after the Consume commit can strand consent, but it cannot cause the queue to release the same decision twice. That is safer than returning first and marking the decision stale afterward.

The public consume API takes the exact Proposal ID and returns the verified Decision plus its recomputed ID inside one lock. Requiring callers to look up a Decision ID first would create a lookup/consume race, while the queue state machine already makes the stored Decision immutable.

The storage boundary is an explicit OCaml module rather than a new builtin or effect. That keeps the 27 kernel forms, `Fs`, `GovernanceApprovalV1`, canonical serialization, and HASH_V0 unchanged. The larger diff remains one refusal boundary: codec, chain verifier, physical commit protocol, locked state transitions, and adversarial tests must agree on the same durable format.

## Safety contract

- Exact Submit and Decide retries are no-ops; configuration drift and conflicting decisions fail closed.
- Approved and Denied require the authenticated actor to match the Decision approver. Escalate retains its host actor as queue evidence without changing released Decision Code.
- Every successful existing-file non-Busy Submit, Decide, Consume, or recovery result syncs the file and parent directory before return, even when the call observes an already committed state without appending. Missing-file recovery has no inode to sync and returns the fixed empty genesis.
- Busy contention performs no mutation.
- Stale state is terminal. Recovery requires a genuinely different Proposal identity.
- The local filesystem contract assumes cooperating advisory-lock users and a trusted, stable parent directory. Pathname checks retain an inherent final TOCTOU window against hostile concurrent renames; no hostile-directory or distributed-filesystem safety is claimed.

## Explicit slice boundary

This is GM.13A, not completion of GM.13. GM.13B still has to add the narrow trusted bridge for the exact released approval operation and prove live gate/Audit ordering under approval, audit-failure, restart, and replay faults. This PR does not authenticate a person, evaluate policy, synthesize consent, invoke a driver, provide a generic host-handler injection API, prove external reviewer identity, or claim exactly-once outside-world action.

Kernel-level deterministic write/fsync fault injection is also not added here. Append and sync errors return E1526 without acknowledging the transition, while the recoverable crash-prefix states are exercised directly. The later fault lane must inject every syscall site without leaving a production test seam.

## How it was checked

- `dune build @all`, formatting, and documentation builds passed.
- All 755 compiled Alcotest/QCheck cases passed, including thirteen GM.13A queue cases.
- All 44 command-line transcript files passed.
- The GM.13A reconstructible overlay manifest verified exactly.
- The complete 0.1 release reproducer passed: runtime memory checks, native differential (69 identical programs and 8 manifested refusals), native leak checks, 1,000 seeded native fuzz programs, installer smoke, packaged demos, CLI, and gauntlet.
- Independent GPT-5.6 Sol xhigh review passed exact commit `83a356450466cb75541162ace9010efb9d201a2b` with no blocking or should-fix findings. The review disposition is attached to the PR.


jbwinters:
## Independent exact-head review

Reviewer: GPT-5.6 Sol xhigh subagent  
Reviewed commit: `83a356450466cb75541162ace9010efb9d201a2b`  
Disposition: **PASS — PR-ready, with no blocking or should-fix findings.**

The first frozen implementation was not accepted. The reviewer found that retries could acknowledge a complete but not-yet-synced commit, that parent-directory durability depended incorrectly on whether the current caller created the file, and that descriptor-close and append-time `fstat` errors could escape the result-valued API. It also required evidence for the real missing-path creation branch, pinned Decide/Decision/Consume identities, committed illegal transitions, and an explicit trusted-parent/TOCTOU limit.

The reviewed commit fixes those findings by establishing file-and-parent barriers for every successful existing-file non-Busy mutating/delivery result, syncing every append through record/commit/file/parent ordering, converting filesystem failures to E1526, and expanding the compiled suite from ten to thirteen GM.13A cases.

Sol independently verified:

- record, commit, file, and parent durability ordering plus retry barriers;
- exact `O_EXCL` creation, `0600` mode, restart, and retry behavior;
- pinned Submit, Decide, Decision, and Consume identities;
- illegal transitions, noncanonical lines, and known-malformed versus unsupported carriers;
- process/Domain concurrency and the explicit trusted stable-parent limitation;
- accurate `755`/13-case evidence, overlay manifest hashes, formatting, and clean exact worktree;
- no persistence, concurrency, codec, evidence, schema, or slice-boundary regression.

Exact-head validation also passed all 755 compiled/property cases, all 44 command-line transcript files, all 27 documentation examples across 8 documents, and the complete 0.1 release reproducer.

PR #54, opened 2026-07-19
Add guarded host bridge for durable approvals

jbwinters:
## Context

GM.13A established a crash-safe, single-use approval queue with exact Submit, Decide, and Consume transitions, but it intentionally stopped short of connecting that queue to a production evaluator route. A host still needed a safe way to intercept the released governance-approval.ask operation, persist the proposal, stop the first invocation, and deliver an authenticated decision only during a later complete rerun.

Without that bridge, a host integration could accidentally resume during the first run if a reviewer raced the Submit, trust mutable Store names instead of the released schema identities, or expose a generic callback that lets in-language code impersonate the production approval boundary.

This PR adds that missing GM.13B host boundary. It keeps the language, kernel, prelude, effect registry, hashes, native runtime, CLI, and release surface unchanged.

## What changed

- Added Governance_approval_bridge.run as an OCaml-only routed driver over the existing evaluator state-capturing boundary.
- Intercepts only the exact released governance-approval.ask effect operation. Every other root operation still goes through the installed-root-handler guard.
- Freezes and verifies the released effect, operation, owning type, and constructor hashes and roles before running. Hostile Store name rebinding fails closed.
- Encodes the exact released GovernanceProposal carrier and decodes only exact Approved, Denied, or Escalate Decision carriers.
- Implements the two-run protocol:
  - a newly Applied Submit always returns Awaiting immediately, even if a reviewer records a Decision concurrently;
  - only an Unchanged Submit from a later complete rerun may Consume;
  - only Delivered resumes the affine continuation and permits the guarded action.
- Admits one approval rendezvous per driver invocation. A second sequential Ask returns the new bridge-specific E1527 diagnostic.
- Preserves typed Busy and stale outcomes and the GM.13A E1520-E1526 queue diagnostics.
- Added eight focused Alcotest cases covering replay, config drift, malformed and second rendezvous input, hostile schema rebinding, Audit ordering, denied and escalated decisions, Busy/corrupt/I/O fail-closed behavior, reviewer contention, and concurrent at-most-once bridge action.
- Added GM.13B release evidence and a reconstructible SHA-256 overlay manifest, and updated successor evidence counts to 763 compiled/property cases.

## Why it was done this way

The bridge deliberately reruns the complete computation instead of persisting an evaluator continuation. Durable queue state can be reconstructed and audited across process restarts; an in-memory affine continuation cannot. Separating Applied from Unchanged is what makes the first invocation safe under a fast-reviewer race: the call that created the durable proposal never consumes or resumes.

The host API is intentionally narrow. It accepts a queue path and authenticated approver configuration, not a generic approval callback or handler registry. That keeps the production trust boundary in the host and prevents a script from installing a handler that grants its own request.

Frozen hashes are checked together with owner declarations and member roles because names in Store are mutable bindings, not sufficient proof of the released protocol identity. E1527 makes a schema mismatch or unsupported second rendezvous distinguishable from malformed Proposal, Decision, queue-state, and I/O failures.

Audit operations in this bridge are handler acknowledgments, not a durable Audit transaction. A later second Ask does not roll back root effects that already occurred, so production hosts must admit computations with exactly one approval rendezvous. The evidence documents these limits rather than implying transactional or exactly-once external actions.

## How it was checked

- The GM.13B SHA-256 manifest verifies every changed file except the self-excluded manifest.
- Build at @all passed.
- Formatting and documentation build gates passed.
- Runtime and readability checks passed.
- All 763 compiled Alcotest/QCheck cases passed in the aggregate ordering that creates Domains before the bridge suite.
- All 44 cram transcripts remain in the frozen successor inventory.
- Full scripts/release/reproduce-0.1.sh passed with the isolated Dune root pinned.
- Native differential testing passed: 69 identical cases, 8 manifested refusals, 0 failures.
- Native leak checks passed.
- Native fuzzing passed 1,000 generated cases with no divergence.
- Installer smoke, packaged demos, CLI/gauntlet lanes, and gauntlet Alcotests passed.
- A fresh independent GPT-5.6 Sol xhigh review of exact commit 8c44e29bbcedcf3fbce33f52daa9b9ff489f5018 returned PASS with no blockers or should-fix findings.

## Explicit exclusions

This PR does not add polling, persisted continuations, multiple approvals in one invocation, a CLI or authentication surface, remote queues, Audit transaction or rollback semantics, exactly-once external actions, or any new language feature.


jbwinters:
Independent review: GPT-5.6 Sol xhigh reviewed exact commit 8c44e29bbcedcf3fbce33f52daa9b9ff489f5018 against bb3f5ead766c914cacd7f7b9caee93ed109aa874 and returned PASS with no blockers or should-fix findings. It explicitly verified Applied-versus-Unchanged delivery, frozen owner/member identities, E1527, guarded root dispatch, Domain-only concurrency coverage, documented Audit and rollback limits, and the complete manifest.

PR #63, opened 2026-07-22
demo: add governed workspace flagship

jbwinters:
## Context

Jacquard already has the individual pieces of a governed effect boundary: a narrow `Workspace` facade, dry-run and live membranes, nested policy attenuation, exact proposal-bound approval, tamper-evident audit logs, and Warp fault exploration. What was missing was one production-shaped narrative showing those pieces governing the same unchanged agent.

Without that synthesis, it was too easy to read the individual examples as unrelated library features. The release needs evidence that an agent can name only its business capability while the surrounding handler composition determines whether calls are simulated, refused, executed through real drivers, held for durable review, or explored under faults.

This PR adds checkout-only demo and release evidence. It does not change the language, kernel, evaluator, native runtime, command-line contract, or packaged release surface.

## What changed

- Added one unchanged `Workspace`-only deployment agent that reads a manifest, fetches an artifact, writes generated configuration, and requests deployment.
- Added a world-free dry run using the released pure simulator membrane. It completes all four calls with an empty outward effect row and zero raw actions.
- Added nested live-policy evidence: a permissive inner membrane forwards the exact call to a stricter outer membrane, which refuses before raw action. Changing only the outer policy lets the same agent complete.
- Added a bounded OCaml evidence host for the actual released `workspace.live-layer`. The live row exposes `{Secret, Fs, Net}` and the transcript pins real driver counts and order; `Secret` exists only in this live driver path.
- Added proposal-only durable review through the released approval bridge and queue. A denial is submitted, recorded, consumed, and validated against the exact deployment proposal while all raw counters remain zero.
- Added a 16-plan `fault.all` world for this exact agent. The immutable four-bit plan is selected before the `Workspace` Once handler and proves every manifest/artifact/write/deployment failure prefix, no later call after failure, no deployment after an earlier failure, and at most one deployment.
- Added runtime-derived call, policy, proposal, and agent identities. The transcript displays both strict and permissive policy IDs and shows the policy-only outcome change with `agent-changed=false`.
- Added three canonical audit-entry data fixtures and verification of the independently supplied final audit head.
- Added sampled and exhaustive Warp laws, a public cram transcript, GM18 release evidence and SHA-256 manifest, and demo/documentation indexes.
- Reconciled the documented cram inventory from 48 to 49 after adding the new transcript.

## Why it was done this way

The agent stays unchanged because the point of an effect membrane is that policy and execution belong outside the governed computation. Its checked row names only `Workspace`; the dry and live handlers make the real authority difference visible.

The durable approval example is deliberately a proposal-only preflight. The released approval bridge persists proposals and decisions, not continuations, and resumes by rerunning the computation. Placing approval after earlier live actions would risk replaying manifest reads, artifact fetches, or writes without a checkpoint, action journal, or idempotent-driver contract. A denied preflight therefore proves zero world actions honestly. This PR does not claim an approved host-orchestration path.

The live host reconstructs the canonical deployment proposal and validates the exact deployment URL and body before incrementing the deployment `Net` counter or entering the provider boundary. The released driver reads and exposes its secret before `Net.fetch`; the evidence says so explicitly. Zero-secret claims come from strict refusal and queue denial, not from request-drift validation.

The agent-specific 16-plan fault world is intentionally bounded to its four facade calls. The existing GM15 349-site/698-path lane remains supporting evidence for the general hostile governance machinery rather than being presented as a substitute for running this agent.

Audit values live under `corpus/governance/` because they are canonical data fixtures, not executable demo programs. This keeps native demo discovery and manifested-refusal accounting truthful.

## Semantic contract and invariants

- The governed agent's outward row is exactly `{Workspace}`; it never receives direct `Fs`, `Net`, or `Secret` authority.
- The dry world is fully handled and performs zero raw actions.
- Nested membranes can only attenuate authority. The inner membrane forwards the exact call unchanged, and the strict outer membrane stops it before raw action.
- The real live leaf exposes every raw capability it may use. Driver counters are host-observed, not simulated-world substitutes.
- Decisions bind the exact canonical proposal ID. Proposal, request, or policy drift fails closed.
- Durable denial occurs before live work and cannot replay world effects.
- Fault choices are fixed before entering Once handlers; a failing call does not resume and no later facade call occurs.
- Audit verification uses the committed canonical entries and an independently supplied expected head.

## Deliberately excluded

- No approved durable-review orchestration path.
- No persisted continuation, checkpoint, action journal, retry protocol, or exactly-once/idempotent-driver guarantee.
- No new language semantics, kernel forms, effects, grants, public CLI behavior, or packaged demo contract.
- No expansion of the frozen native/performance or macro scope.

## How it was checked

Reviewed coordinates:

- Base: `dcdd38457f13b0523ad5e3776e05135c9fa903d7`
- Head: `9539ccda65775482311e4b60064f076eb85af66c`

Exact-head evidence:

- `dune build @all`, `dune build @fmt`, and `dune build @doc` passed.
- The public GM18 launcher and focused cram transcript passed.
- The GM18 SHA-256 evidence manifest verified every listed file.
- Sampled Warp ran 100 policy cases with seed 42; exhaustive Warp covered both policy worlds.
- The exact-agent `fault.all` world covered all 16 immutable plans with prefix distribution 8/4/2/1/1.
- The supporting GM15 lane verified 349 fault sites and 698 typed-error/fail-stop paths.
- The complete compiled/property suite passed all 799 tests.
- All 49 CLI transcripts and 27 documentation examples across 8 documents passed through the full release reproduction.
- Runtime memory, ASAN-oriented leak, interpreter/native differential, and parity gates passed.
- Native differential result: 69 identical programs, 8 manifested refusals, 0 failures.
- Native fuzz passed 1,000 seeded programs with no divergence.
- Installer smoke, packaged demos, case studies, and CLI/gauntlet release evidence passed.
- The mandatory GM12B lane exhaustively verified all 50,000 nested-policy cases: 1 passed, 0 failed, 0 skipped, 0 refused.
- GPT-5.6 Sol xhigh review cycle 1 identified three evidence-claim gaps; all were corrected in one batch. Cycle 2 reviewed this exact head and returned PASS with no blocker or owner decision.

## Risks and follow-up

The OCaml live host is checkout-only release evidence, not a new supported embedding API. It uses fixed deterministic inputs so hashes, audit heads, counters, and ordering remain reviewable. A future approved durable workflow would require a separate design for checkpoints, an action journal, or runtime-enforced idempotent drivers; this PR intentionally does not imply those guarantees.


jbwinters:
## Independent review and final readiness

- Base: `dcdd38457f13b0523ad5e3776e05135c9fa903d7`
- Final/reviewed head: `9539ccda65775482311e4b60064f076eb85af66c`
- Reviewer: GPT-5.6 Sol xhigh
- Final verdict: **PASS**
- Semantic delta since final review: none

Review cycle 1 returned BLOCK on three current-scope evidence issues:

1. The advertised fourth world ran the generic GM15 hostile matrix rather than this exact agent under `fault.all`.
2. The transcript showed strict/permissive outcomes but did not display both runtime policy identities as a semantic policy diff.
3. The deployment-boundary wording could be read as pre-Secret even though the released live driver reads and exposes the secret before `Net.fetch`.

All three findings were accepted and corrected in one batch. The final head now runs the unchanged agent through all 16 immutable four-site fault plans, prints the strict and permissive policy IDs with `agent-changed=false`, and states precisely that request validation precedes the deployment Net/provider boundary while earlier live Fs/Secret work has occurred. GM15 remains supporting infrastructure evidence. No finding was deferred or rejected, and no architecture conflict or owner decision remains.

Final exact-head validation:

- Full compiled/property suite: 799 passed.
- CLI transcripts: 49 passed through release reproduction.
- Documentation examples: 27 across 8 documents passed.
- Runtime memory/parity, native leak, and native differential gates passed.
- Native differential: 69 identical, 8 manifested refusals, 0 failures.
- Native fuzz: 1,000 cases, no divergence.
- Exact-agent fault world: 16 plans, all prefixes valid, deployment at most once.
- Supporting hostile lane: 349 sites / 698 paths.
- GM12B mandatory evidence: 50,000 exhaustive cases, 1 passed, 0 failed/skipped/refused.
- Installer smoke, packaged demos, case studies, CLI/gauntlet, formatting, documentation build, and GM18 manifest passed.

There are no unresolved current-scope blockers or evidence gaps. Required GitHub checks should run on the pushed head. Merge action: squash auto-merge after required checks pass.

PR #74, opened 2026-07-24
Make budget-stopped schedule prefixes strictly replayable

jbwinters:
## Context

A scheduler run can stop because its task budget is exhausted. Jacquard keeps the decisions reached before that stop as a canonical prefix so exhaustive search can inspect or fork it.

The task-limit prefix was canonical but could not be strictly replayed. Recording observed an `async.spawn` decision, validated the attempted child creation, and then discovered that the task budget was already full. The snapshot kept the decision but dropped that in-flight creation. Replaying the saved prefix therefore reached the same spawn and failed with “unexpected creation after trace EOF” instead of reproducing the task-budget stop.

This is a scheduler correctness repair. It does not change Jacquard source syntax, evaluator results, scheduling choices, trace format, or release identity.

## What changed

- Split task allocation into a budget preflight and a mutation step.
- Apply the same preflight to child spawns and nested-scope root creation.
- Refuse an over-budget creation before recording a `Create` event or mutating task counts.
- Continue validating allowed creations before allocation state changes.
- Make prefix snapshots retain any creations that were already validated during an in-flight decision, in canonical `Decide` then `Create` order.
- Add an end-to-end regression that records a real task-budget stop, strictly replays it with the same bounds, and requires the same diagnostic and byte-identical prefix.
- Update the mechanically checked successor test inventory from 826 to 827 and the schedule-trace group count from 3 to 4.

## Why it was done this way

A trace whose header says `max-tasks=1` cannot legally contain a second task creation. Adding the refused child to the prefix would make the trace fail its own canonical validator.

The scheduler instead checks the same task limit at the same point in record and replay modes. Both runs now observe the spawn, refuse it before creation, and return the same canonical stopped prefix. Allowed creations still keep the stronger replay rule: validate the exact expected creation before mutating allocation state.

The snapshot helper also now preserves genuinely validated in-flight creations, so its partial trace uses the same event order as a completed decision.

## Semantic contract

- An over-budget spawn or nested scope records no creation and allocates nothing.
- An allowed creation is replay-validated before allocation mutation.
- Committed event order remains `Decide -> Create`.
- Strict replay with identical bounds reproduces the same task-limit refusal and byte-identical prefix.
- Existing fork ordering and exhaustive-search behavior remain unchanged.

## Deliberately excluded

- No schedule format or version change.
- No `HASH_V0`, canonical program identity, or cache-key change.
- No scheduler-choice, task-bound, evaluator, kernel, `.jac`, or `.jqd` semantic change.
- No native-runtime behavior change.
- No historical release manifest bytes were rewritten.

## How it was checked

- Regression-first failure reproduced the old drift: `unexpected creation of 0#1 after trace EOF`.
- Focused suites passed: 4 schedule-trace cases, 1 round-robin case, and 8 exhaustive-schedule cases.
- Final forced full suite passed all 827 Alcotest/QCheck cases in 339.790 seconds, plus all cram transcripts and 27 doctest examples.
- Forced GM12B forwarding evidence passed 50,000 exhaustive cases with zero failures, skips, or refusals.
- Full build, documentation build, runtime checks, formatting, whitespace, version smoke, and installer smoke passed.
- The production historical-publication gate verified all 35 registered publications and rejected manifest drift, missing/unregistered entries, coordinated deletion, and checker-policy weakening.
- GPT-5.6 Sol xhigh independently reviewed exact head `123d0694d617b69d2038e2d9dbbc8de36cfbf6cb` against base `3522906a50e2d83bb5783d829a9262f3d159deab` and returned **PASS** with no blocking or nonblocking findings.

## Risks and follow-ups

The regression directly covers the `async.spawn` task-limit path. Nested `async.scope` uses the same preflight implementation and its existing global-bound tests remain green, but it does not yet have a symmetric stopped-prefix replay case. The private controller’s in-flight creation snapshot behavior is straightforward and reviewed, but it does not have a direct controller-level unit test because that module is intentionally outside the public test boundary. These are evidence-depth follow-ups, not known behavior defects.

### DIFF
--- a/docs/release/dx-jac-export/EVIDENCE.md
+++ b/docs/release/dx-jac-export/EVIDENCE.md
@@ -6,7 +6,7 @@ artifacts and are not retroactively updated by DX.2.
 
 ## Current inventory
 
-- Alcotest/QCheck cases: `826`
+- Alcotest/QCheck cases: `827`
 - Cram transcript files: `51`
 - Doctest examples: `27` across 8 documents
 
--- a/docs/release/structured-concurrency/EVIDENCE.md
+++ b/docs/release/structured-concurrency/EVIDENCE.md
@@ -34,7 +34,7 @@ native Channel, actor, or supervision claim.
 | C3 | Scoped typed channels run through deterministic FIFO, seeded, replay, exhaustive, and cached interpreter scheduling with exact run/scope ownership, rendezvous and buffering, close, cancellation, and deadlock behavior. | `channel-contract`, `round-robin`, and `exhaustive-schedule` suites; `test/cli/task-values.t`; `test/cli/schedule-replay.t`; and the frozen traces below |
 | C4 | Not claimed: host asynchronous I/O, actors, and supervision are absent. | [LIMITS.md](LIMITS.md) |
 
-The current successor inventory is exactly 826 compiled Alcotest/QCheck cases, 51 recursive
+The current successor inventory is exactly 827 compiled Alcotest/QCheck cases, 51 recursive
 cram transcript files, and 27 named doctest examples across 8 documents. The
 repository release-law checks recompute those counts instead of trusting this
 paragraph.
@@ -738,7 +738,7 @@ eight independently selectable groups and their case counts are:
 | `cancellation` | 1 | cooperative boundary delivery |
 | `scope-policy` | 1 | fail-fast and collect aggregation |
 | `round-robin` | 1 | real evaluator FIFO lifecycle, including Channel seeded/replay/cache parity |
-| `schedule-trace` | 3 | canonical identity; refusal compatibility; impossible-event refusal |
+| `schedule-trace` | 4 | canonical identity; refusal compatibility; impossible-event refusal; task-budget prefix replay |
 | `exhaustive-schedule` | 8 | hand counts and Channel worlds; Warp/replay; failures; budgets; hermeticity; Once ownership |
 
 The exact discovery and focused execution commands are:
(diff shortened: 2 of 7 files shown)


jbwinters:
## Independent review, finding disposition, and final readiness

**Reviewer:** GPT-5.6 Sol xhigh  
**Base:** `3522906a50e2d83bb5783d829a9262f3d159deab`  
**Final reviewed head:** `123d0694d617b69d2038e2d9dbbc8de36cfbf6cb`  
**Verdict:** **PASS**

### Findings

- **Accepted current-scope evidence correction:** the structured-concurrency discovery table still listed 3 schedule-trace cases after the new regression raised that group to 4. The final head updates the count and names task-budget prefix replay. Focused surface-law validation and the final full suite passed afterward.
- **Remaining blocking findings:** none.
- **Remaining nonblocking findings:** none.
- **Architecture conflicts or owner decisions:** none.
- **Disputed reruns:** none.

### Evidence gaps classified as test-depth debt

- The regression directly covers task-limit replay for `async.spawn`; nested `async.scope` uses the same production preflight but lacks a symmetric replay case.
- The private schedule controller’s in-flight creation retention has no direct controller-level test because that module remains outside the public test boundary.

Neither gap indicates a known behavior defect or changes the reviewed contract.

### Final readiness

- Exact final head: `123d0694d617b69d2038e2d9dbbc8de36cfbf6cb`; worktree clean and directly based on the stated base.
- Focused schedule-trace, round-robin, exhaustive-schedule, and surface-law suites passed.
- Exact-head forced full suite: 827 tests passed in 339.790 seconds; all cram and doctest checks passed.
- Forced GM12B evidence: 50,000 exhaustive cases passed; 0 failed, skipped, or refused.
- Build, docs, runtime checks, formatting, whitespace, installer/version smoke, temporary-storage policy, and the 35-publication historical mutation gate passed.
- The final head is the exact reviewed head; semantic delta after review: none.
- GitHub protected checks are pending. Squash auto-merge will be enabled; no unresolved current-scope debt remains.

PR #75, opened 2026-07-24
Make channel deadlock diagnostics describe the real trigger

jbwinters:
## Context

Jacquard reports E0908 when the deterministic scheduler has no runnable task
and the scope cannot make progress because channel work is involved. The old
message said every remaining task was blocked on a channel.

That was narrower than the real rule. The scheduler also reports this state
when one task is waiting on a channel and the other live tasks are suspended
for another reason, such as waiting for a task or a nested scope. The behavior
was correct, but the diagnostic and current documentation could mislead
someone debugging a mixed wait.

## What changed

- Reworded the E0908 message to say: `all remaining live tasks are suspended
  and at least one is blocked on a channel`.
- Updated the normative concurrency guide and current successor evidence to
  state the same condition.
- Updated the exact-message regression.
- Added a direct regression with one channel receiver and one yielded task. It
  proves that the mixed suspended state is deadlocked and that making the
  yielded task runnable clears the predicate.
- Updated the documentation assertion that keeps the release checklist in
  sync.

## Why it was done this way

The runtime predicate already matched the intended scheduler contract, so this
change corrects the explanation instead of changing scheduling behavior. The
new wording tells a reader both facts the runtime actually knows: every live
task is suspended, and at least one suspension is a channel wait.

The mixed-state regression is at the structured-scope boundary where the
predicate is observable. The existing end-to-end scheduler regression still
pins the exact public E0908 message.

## Semantic contract

- E0908 is reported at an empty runnable queue only when the scope is open,
  at least one live task remains, every live task is suspended, and at least
  one task is waiting to send or receive on a channel.
- Other suspended tasks may be waiting for tasks, yielded for a nested scope,
  or waiting for host readiness.
- The diagnostic code, domain, span behavior, cleanup, task ordering, scope
  policy, and deadlock predicate are unchanged.

## Deliberately excluded

- No scheduler, channel, cleanup, or policy behavior change.
- No kernel, surface syntax, prelude, native runtime, trace, cache, or hash
  change.
- No rewrite of historical release manifests or historical evidence bytes.
- No new corpus program because the only public behavior change is stable
  diagnostic text, pinned directly by unit coverage.

## How it was checked

- The exact-message regression failed first against the old runtime text.
- Focused `round-robin`, `structured-scope`, and `effect-taxonomy` suites
  passed after the correction.
- Fresh GPT-5.6 Sol xhigh independent review passed exact commit
  `52f9c154de7ea4eb4f256239222d9ff7e551a093` with no blockers or architecture
  conflict.
- `dune build @all @doc` passed.
- The forced full suite passed all 827 compiled/property tests in 344.342
  seconds, plus every command-line transcript and all 27 documentation
  examples.
- Temporary-storage policy, installer smoke, formatting cleanliness, version
  `0.1.0`, release-file presence, and whitespace checks passed.
- All 35 historical publications and the complete manifest mutation battery
  passed.
- The forced GM12B forwarding proof passed all 50,000 cases with zero
  failures, skips, or refusals.

## Risks and follow-ups

The changed diagnostic text is intentional and may affect consumers that
match the full message instead of the stable E0908 code. The exact public
message is now pinned.

Additional direct cases for every suspension variant would add evidence, but
the unchanged predicate is already covered by the mixed-yield regression and
was independently reviewed. No correctness follow-up is required for this
slice.

### DIFF
--- a/docs/concurrency.md
+++ b/docs/concurrency.md
@@ -968,11 +968,12 @@ results, or deadlock detection. Jacquard scopes select fail-fast:
 - Collect does not cancel siblings or implicitly close a channel when one child
   fails. Other children may continue to communicate and their results remain in
   creation order;
-- under either policy, if every remaining live task is channel-blocked and no
-  channel transition is possible, the scheduler reports deadlock E0908. In
-  particular, fail-fast has no failure to prefer in this state; Collect cannot
-  return a partial aggregate. Deadlock never implies an implicit close. Cleanup
-  removes waiters and destroys continuations;
+- under either policy, if every remaining live task is suspended and at least
+  one is channel-blocked, the scheduler reports deadlock E0908. The other
+  suspended tasks may be waiting for tasks or nested scopes. In particular,
+  fail-fast has no failure to prefer in this state; Collect cannot return a
+  partial aggregate. Deadlock never implies an implicit close. Cleanup removes
+  waiters and destroys continuations;
 - explicit close while the scope is active uses §8.3 regardless of policy;
 - final scope destruction closes owned channels in creation order, removes any
   waiter before destroying its Once continuation, and drops undrained buffered
@@ -999,8 +1000,9 @@ SC.14 is conforming only if all of the following remain true:
   double continuation consumption;
 - [x] exact run/scope ownership, recursive escape scans, hostile callback
   revalidation, and deterministic ChannelId creation;
-- [x] fail-fast cancellation, OCaml Collect non-interference, policy-independent
-  all-channel-blocked E0908 refusal, and exception-safe teardown;
+- [x] fail-fast cancellation, OCaml Collect non-interference, the
+  policy-independent all-suspended/at-least-one-channel-blocked E0908 refusal,
+  and exception-safe teardown;
 - [x] rendezvous and buffered contract traces plus positive/negative checker
   fixtures.
 
--- a/docs/release/structured-concurrency/EVIDENCE.md
+++ b/docs/release/structured-concurrency/EVIDENCE.md
@@ -698,8 +698,9 @@ idempotent drain-on-close. Cancellation is delivered before channel mutation
 and removes a blocked waiter without reordering survivors. Handles are exact
 run/scope capabilities and escape or parent/descendant use is E0907.
 Fail-fast removes channel-blocked siblings through cancellation; collect does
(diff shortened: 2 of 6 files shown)


jbwinters:
## Independent review and final readiness

- Reviewer: GPT-5.6 Sol xhigh, fresh independent subagent
- Reviewed/final head:
  `52f9c154de7ea4eb4f256239222d9ff7e551a093`
- Base: `89692ba7fd0f34097f07f843c822b301ea793cd3`
- Verdict: **PASS**
- Blocking findings: none
- Architecture conflicts: none
- Disputed reruns: none
- Semantic delta after review: none

The reviewer confirmed that the unchanged scope predicate requires an open
scope, a nonempty live set, every live task suspended, and at least one channel
send/receive wait. The runtime change affects only the text emitted at that
existing empty-queue boundary. E0908, its diagnostic domain/span behavior,
cleanup, and scheduling remain unchanged.

### Finding disposition

- **Accepted current-scope corrections:** none; the exact reviewed diff passed.
- **Deferred correctness debt:** none.
- **Rejected findings:** none.
- **Architecture decisions:** none.
- **Optional evidence gaps:** direct predicate cases do not separately cover
  every suspension variant, and the end-to-end scheduler case is channel-only.
  The direct regression covers the important mixed Channel + Yielded state,
  while the unchanged predicate and public-interface contract cover the other
  variants. No current-scope edit is warranted.
- A local user Git setting treats ordinary OCaml space indentation as a
  whitespace warning. The repository-standard trailing-space and
  space-before-tab check passes; formatting is clean.

### Final exact-head validation

- Regression-first exact-message failure captured against the old runtime text.
- Focused `round-robin`, `structured-scope`, and `effect-taxonomy` suites:
  pass.
- `dune build @all @doc`: pass.
- Forced full suite: all 827 tests passed in 344.342 seconds; every cram
  transcript and all 27 documentation examples passed.
- Historical gate: all 35 publications plus the full mutation battery passed.
- Installer smoke, temporary-storage policy, formatting/clean diff, whitespace,
  version `0.1.0`, and release-file presence: pass.
- Forced GM12B proof: 50,000 cases passed; zero failures, skips, or refusals.
- Remaining debt: optional test variants only; no correctness blocker.
- Merge action: squash auto-merge is enabled and will wait for protected
  GitHub checks.

PR #82, opened 2026-07-25
Pin formatter width boundaries and readable continuations

jbwinters:
## Context

The canonical `.jac` formatter did not have a reliable contract at its 100-byte boundary. OCaml's formatting engine could break a group that fit exactly, and the fallback search for a narrower internal margin could remove indentation from a signature or labeled field. That made a one-byte edit change the visual structure of code in a misleading way.

Two valid but unbreakable forms also needed an honest policy. Type and effect declaration headers must keep their name with `=` or `where`, and a `forall` prefix has no grammar continuation in 0.1. The earlier documentation overcorrected by claiming that every other breakable line always fits. Concrete match-clause and raw-kernel examples show that the width is a layout target, not a global physical-line guarantee.

This PR completes the formatter boundary work without changing the language grammar, kernel, lowering, evaluator, native runtime, or file-carrier contract.

## What changed

- Measures complete formatting groups before choosing compact or multiline layout.
- Keeps groups compact when they fit exactly at the requested width.
- Uses one-item-per-line layout and trailing commas for multiline comma-separated groups.
- Preserves stable append/remove diffs for multiline lists.
- Keeps `else if` chains flat and adds W1203 for unusually large match scrutinees.
- Breaks top-level signatures and labeled constructor fields after `:`, with the colon on the owning line and a structural two-space type continuation.
- Removes the descending narrow-margin search that could erase continuation indentation.
- Adds W1204 for type/effect declaration headers whose shortest legal form exceeds 100 bytes.
- Adds W1205 for every over-wide `forall` prefix, including prefixes inside expression annotations, with diagnostics on the exact prefix in source order.
- Documents UTF-8 byte measurement and the three deliberate unbreakable forms.
- Narrows the overall promise: 100 bytes is a layout target, exact-fit behavior is decided per top-level item, and known overflow cases are explicitly non-exhaustive.
- Adds exact boundary, idempotence, lowering, hash-invariance, diff-stability, diagnostic-span, and public CLI coverage.

## Why it was done this way

The formatter is review infrastructure. Stable layout lets a reader see structure and meaning without mentally correcting indentation or interpreting avoidable diff churn. The exact-fit preflight keeps compact code compact, while falling back only to the requested margin preserves nesting cues when a line cannot fit.

The two grammar-driven exceptions remain valid `.jac` instead of being rejected or silently moved to `.jqd`. W1204 and W1205 make those cases visible and actionable. Multiline declaration headers and quantified binders remain a post-0.1 grammar question, which keeps this release-hardening change out of the language and kernel.

The documentation describes the behavior that is actually verified. It does not turn a formatter target into a stronger global guarantee than the implementation can support.

## How it was checked

- Full build passed.
- All 838 compiled and property tests passed in 374.391 seconds.
- Public command-line formatter transcripts passed.
- Packaged demo and Linux x86-64 installer smoke tests passed.
- Native/runtime checks passed.
- Documentation build and formatting checks passed.
- Historical evidence checks passed, including mutation tests that reject manifest drift.
- The immutable SS.22 surface-syntax evidence manifest remains unchanged.
- Version smoke reported `0.1.0`.
- Trailing-space, whitespace, and clean-worktree checks passed.
- Claude Opus 5 independently reviewed exact commit `555795662e78762f9261395f816c8f6d85d52d64` and returned PASS after reproducing the boundary and documented-overflow cases.

## Scope

There is no grammar, kernel-form, lowering, canonical-hash, evaluator, native-runtime, or carrier-format change. Parser additions are hash-excluded source metadata used only for precise diagnostics. Readability-study results are not a completion gate for this existing surface-syntax work.

### DIFF
--- a/docs/errors.md
+++ b/docs/errors.md
@@ -234,6 +234,7 @@ also emits E0817; consuming the captured resumption twice emits E0816.
 | W1201 | lowercase binding pattern shadows an in-scope constructor differing only in case | `match Up { | up -> ... }` |
 | W1202 | positional constructor pattern has more than four fields | `Snapshot(_, _, _, _, _)` |
 | W1203 | match scrutinee spans more than four source lines | manually bind the expression with `let`, then match on its name |
+| W1204 | shortest legal type/effect declaration header exceeds the canonical formatter width | shorten the declaration name or type-variable list |
 
 ## Explicit bootstrap export (E13xx)
 
--- a/docs/errors.md
+++ b/docs/errors.md
@@ -235,6 +235,7 @@ also emits E0817; consuming the captured resumption twice emits E0816.
 | W1202 | positional constructor pattern has more than four fields | `Snapshot(_, _, _, _, _)` |
 | W1203 | match scrutinee spans more than four source lines | manually bind the expression with `let`, then match on its name |
 | W1204 | shortest legal type/effect declaration header exceeds the canonical formatter width | shorten the declaration name or type-variable list |
+| W1205 | shortest legal `forall` prefix exceeds the canonical formatter width | split the declaration or reduce its quantified variables |
 
 ## Explicit bootstrap export (E13xx)
 
--- a/docs/release/dx-jac-export/EVIDENCE.md
+++ b/docs/release/dx-jac-export/EVIDENCE.md
@@ -6,7 +6,7 @@ artifacts and are not retroactively updated by DX.2.
 
 ## Current inventory
 
-- Alcotest/QCheck cases: `832`
+- Alcotest/QCheck cases: `835`
 - Cram transcript files: `51`
 - Doctest examples: `27` across 8 documents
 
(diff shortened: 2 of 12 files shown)


jbwinters:
Independent review: PASS

Reviewer: Claude Opus 5, xhigh effort
Reviewed head: 555795662e78762f9261395f816c8f6d85d52d64
Base: d8eeb9867b12070bee5e71db52f51fc89c6c276a

The reviewer independently reproduced:

- the per-top-level width+1 preflight and requested-width fallback;
- exact-fit subgroup breaking when another line forces fallback;
- match-clause arrow and raw jqd overflow cases;
- W1204 and W1205 as necessary rather than exhaustive warnings;
- signature and labeled-field continuation indentation;
- formatter idempotence on the documented overflow cases;
- the docs-only scope of the final contract correction.

No blockers or owner decisions remained. The reviewer found the exact head safe to push and open for squash auto-merge.

PR #81, opened 2026-07-25
Accept trailing commas and stabilize multiline formatting

jbwinters:
## Context

Jacquard already had a canonical 100-column formatter, but comma-separated surface forms still behaved inconsistently. Calls, lists, rows, labeled fields, and several parameter forms rejected a final comma, while wrapped output did not consistently put one item on each line with a diff-stable final comma. That made ordinary edits noisier and left the documented formatter contract incomplete.

There was also a correctness risk around comments next to a final comma: the formatter must preserve the comment without adding blank lines, losing the comma required by a closed multiline row, or moving the closing delimiter out of alignment.

This PR completes that surface-formatting contract. It changes accepted `.jac` punctuation and canonical `.jac` layout only. It does not change the kernel, evaluation, hashes, `.jqd`, the runtime, or native compilation.

## What changed

- Accepts one optional final comma in calls, lists, expression and type tuples, function and handler parameter lists, arrow parameter lists, constructor patterns, labeled constructor fields, operation parameter lists, and effect lists.
- Uses one canonical comma-list layout:
  - compact groups omit the final comma, except singleton tuples where the comma carries arity;
  - wrapped groups put one item on each indented line, end every item with a comma, and put the closing delimiter on its own aligned line.
- Makes closed multiline effect rows end with a comma. Open rows omit the comma before `|` because the row tail continues the list.
- Preserves comments attached to a final comma across every supported context. The printer owns the comment breaks and closing alignment, so formatting does not insert an empty line or indent the closing delimiter incorrectly.
- Extends row trivia ownership to include the possible final comma, allowing comment-bearing commas to survive parsing and printing without affecting semantic identity.
- Updates the surface grammar, formatter documentation, and agent onboarding guidance.
- Replaces obsolete trailing-comma rejection assertions with acceptance, recovery, idempotence, comment-ownership, layout, and hash-inertia coverage.
- Updates the live successor test inventories from 829 to 832. Frozen 0.1 counts and historical manifests are unchanged.

## Why it was done this way

A trailing comma is useful editing punctuation, but it should not become a new semantic distinction. The parser therefore accepts it as surface metadata and lowering produces the same existing kernel forms as the spelling without it. Canonical hashes, evaluation order, and `.jqd` behavior remain identical.

[...]

PR #84, opened 2026-07-26
Add readable marked text interpolation

jbwinters:
## Context

Jacquard already had a variadic `text.join`, but ordinary messages were still harder to read than
they needed to be. Reviewers had to visually reconstruct the boundaries between literal text,
explicit conversions, and joined values. At the same time, implicit rendering would hide type and
authority decisions that the language is designed to keep visible.

This PR adds an opt-in marked text form that improves the common reading path without changing
plain strings, the 27-form kernel, evaluation order, or HASH_V0 identity.

## What changed

- Added `$"total: {text.from-int(count)}"` syntax with `{{` for a literal opening brace and a bare
  `}` treated as text.
- Kept interpolation line-bounded so malformed input recovers at the next source line with one
  opener-anchored E1219 diagnostic.
- Rejected nested marked interpolation in 0.1 with guidance to bind the inner text first.
- Lowered every marked form locally to one ordinary resolved variadic `text.join` application.
- Required every embedded expression to have type `Text`; other values receive explicit
  `text.from-int` conversion guidance.
- Preserved exactly-once, left-to-right evaluation so visible effects follow source order.
- Added canonical marked printing that stays on one physical line even at narrow formatter widths.
  Structurally multiline blocks, matches, or handlers fall back to the equivalent explicit
  `text.join` spelling so formatter output remains valid and idempotent.
- Preserved interpolation provenance through quote/unquote handling and guarded the printer against
  incomplete or corrupted callee metadata.
- Added parser, lexer, checker, formatter, hash-parity, quote, CLI, corpus-twin, doctest, recovery,
  CRLF, and effect-order evidence.
- Updated the public syntax guide, standard-library documentation, agent onboarding skill, error
  registry, and live successor evidence inventories. The immutable SS.22 manifest remains
  unchanged.
- Regenerated the generator-owned `stdlib-ss22.jac` twin, picking up the canonical comma layout
  already shipped by the formatter work in #81/#82; this is mechanical formatting drift, not an
  interpolation semantic change.

## Why it was done this way

The marked `$"..."` prefix makes the feature explicit and leaves every existing plain string and
hash unchanged. Requiring `Text` rather than invoking an implicit `Show` keeps conversions visible
to readers and avoids adding hidden behavior or authority.

Lowering directly to `text.join` gives the readability benefit without creating a second runtime
semantics or expanding the frozen kernel. The single-line rule keeps lexer recovery local, avoids
turning indentation into string data, and prevents a later tightening from becoming a compatibility
break. The explicit-call formatter fallback handles structurally multiline expressions without ever
emitting marked syntax that the lexer would reject.

The syntax is recorded as D75 because D45 is already the global handler-multiplicity decision.
Release prose distinguishes the historical SS.22 boundary from the later SX.26 successor feature
instead of rewriting historical evidence.

## How it was checked

- `opam exec -- dune build @all` passed.
- `opam exec -- dune runtest --force` passed all 843 compiled/property tests.
- All 51 command-line transcript files passed.
- All 28 documentation examples across 8 documents passed.
- Surface twin generation passed for 25 pairs, including the curated interpolation parity fixture.
- Golden generation produced the expected 50 identity rows.
- Formatter idempotence, narrow-width output, multiline fallback, CRLF/EOF recovery, exact
  diagnostic spans, explicit conversion, quote/unquote, effect order, and HASH_V0 parity are pinned
  by focused tests.
- `opam exec -- dune build @doc` passed.
- Formatting and whitespace checks passed.
- The immutable SS.22 manifest check passed against boundary commit `52f36133`.
- Full `scripts/release/reproduce-0.1.sh` passed, including 70 native differential programs, 8
  expected native refusals, native leak checks, 1,000 deterministic native fuzz cases, installer
  smoke, demos, Warp proofs, and gauntlet tests.
- Claude Opus 5 at xhigh effort independently reviewed exact commit `f331c9b` and returned PASS
  after two rounds of concrete findings were fixed.


jbwinters:
Independent review and final readiness: PASS

- Reviewer: Claude Code, Opus 5, xhigh effort
- Intended base: `98b7a7e7a01fd89ed3079f58c41f932af8062861`
- Reviewed and final branch head: `f331c9b8fb44eb9bdea5b5515834c573df4a5161`
- Squash merge: `cf5af44cf1fb8a38707fb310ad7a898e0c42bdcd`
- Tree identity: the reviewed head and squash merge both have tree `8f97d36688b765d4218c283832c10472791f7532`

The final verdict was `PASS`. Earlier current-scope findings were accepted and fixed before the reviewed head: the global decision identifier was corrected from the already-used D45 to D75; successor release wording was separated from the historical SS.22 claim; the formatter's explicit-`text.join` fallback for structurally multiline expressions was documented and tested; and the agent-onboarding grammar was brought into agreement with the shipped marked form. No current-scope blocker, architecture conflict, disputed rerun, or unresolved evidence gap remained.

Preference-only observations were not treated as blockers: repeating the nested-interpolation restriction in one additional onboarding section, wrapping one 84-column prose line, and restating the bare-`}` rule more explicitly there. They do not change behavior or evidence and require no deferred-debt issue.

Final exact-head evidence included `dune build @all`, `dune build @doc`, formatter cleanliness, all 843 compiled/property tests, all 51 CLI transcripts, 28 documentation examples, 25 semantic corpus twins and 50 golden identity rows, plus the full 0.1 reproduction (70 native differential programs, 8 expected refusals, leak checks, 1,000 seeded fuzz cases, installer smoke, demos, Warp, and gauntlet). The immutable SS.22 manifest remained byte-identical.

GitHub then passed Development, clang parity, GCC parity, Governance, and GM12B on the exact branch head and auto-merged with `SQUASH`. The same five required contexts passed again on exact merged `main`. There is no semantic delta after review and no remaining task-owned open PR.

Issue #86, opened 2026-08-09
UX.1: admit authorized readability results and reproduce analysis


jbwinters:
## Context

`readability-protocol-v1` deliberately stops at deterministic planning and
rejects every real human or model result. Before authorized collection can
open, the result-admission and analysis contracts need their own reviewable
engineering slices.

This issue records that already-planned UX.1 work and the nonblocking findings
from the independent review of the planning contract. It does not authorize a
study or expand Jacquard language behavior.

## Result-admission contract

- Require an approved authority manifest and an attested, quota-sufficient
  cohort manifest where applicable.
- Bind every admitted row to the exact frozen human or cohort schedule,
  fixture/prompt/schema digests, carrier, job, ordinal, repetition, run kind,
  and fresh-session settings.
- Preserve failed attempt 1 rows and allow only the preregistered retry shape;
  reject duplicate, missing, reordered, substituted, or cross-cohort trials.
- Harmonize `consent_version` documentation with the human-row schema while
  keeping consent/contact records outside answer data.
- Reserve `__timeout__`, `__system_failure__`, and `__parse_failure__` so a
  fixture option cannot shadow a failure sentinel.
- Reject empty or incomplete real stores and validate the full Draft 2020-12
  human/model branches plus cross-row invariants.
- Clarify that arbitrary-seed `assign` output is a debugging aid, or restrict
  it to the frozen protocol seed; admitted rows must always match the checked
  schedule.

## Analysis contract

- Restate the prior accuracy sample-size premise so the v1 preregistration is
  self-contained.
- Produce separate deterministic outputs for perceived readability,
  comprehension, review/defect detection, modification/debugging, and
  diagnostic recovery, including time, accuracy, confidence, calibration,
  effect sizes, uncertainty, exclusions, expertise, and learning effects.
- Run twice from clean stores and require byte-identical tables/statistics
  except for explicitly declared timestamps.
- Support blinded independent rescoring and prove every published number comes
  from checked, de-identified evidence.

## Exclusions

No model contact, participant recruitment, real collection, approval claim,
readability claim, language/runtime change, or release claim is authorized by
this issue. External consent, ethics/privacy, compensation, accessibility,
retention, publication-license, and data-governance authority remain mandatory
before collection.

## Completion evidence

- adversarial row/store mutation tests for every invariant above;
- deterministic synthetic dry runs confined to `.scratch/` and marked
  non-citable;
- exact-head local and required GitHub gates;
[...]

PR #90, opened 2026-08-10
Add deterministic readability analysis admission

jbwinters:
## Context

Issue #86 can already reject incomplete or unauthorized readability result stores, but accepted rows still need one reproducible rule for deciding which subjects and trials may enter later statistics. In particular, a failed retry or a second system failure must exclude the whole human subject rather than only the failed row.

## What changed

- Add a `prepare-analysis` command that validates the complete store first, then emits canonical `readability-analysis-input-v1` JSON.
- Bind that bundle to the exact source-file SHA-256, ordered row identities, and the reviewed schema, fixture, authority, assignment, schedule, and cohort pins.
- Publish per-subject source, effective, and excluded row IDs plus total and per-carrier flow counts.
- Derive the frozen human retry and whole-subject exclusion transitions; keep model sessions row-local and synthetic output explicitly non-citable.
- Require row-level system/parse failure exclusions to match the deterministically scored failure outcome in both directions.
- Make the checked Dune alias generate two analysis bundles and require byte-identical output.

## Semantic contract

- No system failure: use the human's five first attempts.
- One failed first attempt: preserve that row and use the final retry if the retry itself does not have another system failure. An incorrect answer or timeout remains an outcome.
- More than one verified system failure, including a failed retry: exclude the entire human subject.
- Any stable non-system human exclusion: exclude the entire human subject.
- A model session is effective exactly when its admitted row is not excluded.
- Every source row belongs to exactly one effective or excluded list. The bundle has no timestamp, measured outcome statistic, or readability claim.

## Deliberate exclusions

This does not contact a model, recruit participants, collect real outcomes, claim approval or consent, compute inferential tables, perform blinded rescoring, or make a readability, language, runtime, or release claim. Pre-assignment consent and eligibility flow remains `external-required` because approved enrollment records live outside answer JSONL; this change does not invent a data-governance schema.

## Validation

- Fresh Fable 5 review of `9a3801794e6388c3d5ad3cb1afd9ee1e303ce7a5`: `PASS`, no blockers or architecture conflicts.
- Readability protocol verification: 5 jobs, 3 carriers, 15 dry-run conditions, 480 human assignments, 450 M0 model trials.
- Adversarial self-tests cover a successful retry, failed retry, multiple first-attempt failures, stable subject exclusion, excluded model session, strict failure labeling, complete flow partitions, and exact-source digest sensitivity.
- Two independent synthetic bundles were byte-identical (`53c44754d4203e37086849cc945ef4e26988a39437a6fc3068de4b74ec3fdb9d`) from source `ecc5acb96f439f3a2f39c9bfdc6dde8e174ab315d17ceee7728080342cf5db01`.
- `dune build @all`, `dune build @doc`, `dune fmt`, historical-manifest verification and mutation tests, version smoke, whitespace check, and clean-tree check passed.
- All 847 compiled Alcotest/QCheck tests passed. Existing native sanitizer crams cannot run under the Codex ptrace boundary; an outside-sandbox retry reproduced that explicit LeakSanitizer limitation while the direct runtime suite reported `runtime check: PASS`. Required GitHub clang and GCC lanes remain the clean-host sanitizer evidence.

## Risk and follow-up

The analysis and benchmark modules duplicate a few frozen version/schedule constants. Tests cross-check those constants against each other and against generated schedule bytes, so this is optional cleanup rather than a correctness blocker. Later #86 slices still own inferential outputs, uncertainty/effect sizes, blinded rescoring, published-number derivation, and measured claim gates.

Part of #86.

### DIFF
--- a/docs/research/readability-benchmark/EXECUTION.md
+++ b/docs/research/readability-benchmark/EXECUTION.md
@@ -163,6 +163,46 @@ Preserve, under the approved data controls:
 No human-identifying data, provider secrets, raw model conversations, or
 synthetic rows represented as outcomes may be committed.
 
+## Deterministic analysis input
+
+After `validate-results` accepts a complete store, prepare its versioned row
+selection and provenance bundle with the same exact admission context:
+
+```text
+python3 test/readability/readability_benchmark.py prepare-analysis \
+  --manifest test/readability/fixture-manifest.json \
+  --schema test/readability/result.schema.json \
+  --authority AUTHORITY.json \
+  --input HUMAN-RESULTS.jsonl \
+  > .scratch/readability/human-analysis-input.json
+
+python3 test/readability/readability_benchmark.py prepare-analysis \
+  --manifest test/readability/fixture-manifest.json \
+  --schema test/readability/result.schema.json \
+  --authority AUTHORITY.json --cohort COHORT.json \
+  --input MODEL-RESULTS.jsonl \
+  > .scratch/readability/model-analysis-input.json
+```
+
+The command validates the complete store before emitting anything. Its
+canonical `readability-analysis-input-v1` JSON binds the exact source-file
+SHA-256, ordered row identity, reviewed pins, per-subject decisions, and total
+and per-carrier flow. A successful human retry replaces only its failed first
+attempt; successful here means the retry itself did not have a system failure,
+not that its answer was correct. More than one verified system failure,
+including a failed retry, or
+any stable non-system exclusion removes the whole human subject from effective
+rows. Model sessions remain row-local. Source, effective, and excluded row-ID
+lists partition the input so every count can be recomputed.
+
+The output deliberately omits timestamps and measured outcomes. Running the
+command twice on byte-identical input therefore produces byte-identical output.
+Human pre-assignment consent and eligibility flow is `external-required`
+because those approved records remain outside answer JSONL. This tool neither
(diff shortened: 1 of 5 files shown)


jbwinters:
## Independent review, finding disposition, and final readiness

- Reviewer: Claude Code using Fable 5
- Base: `9b7a1521a71198dcc8a2d755a4d539b6c378bff2`
- Reviewed and final head: `9a3801794e6388c3d5ad3cb1afd9ee1e303ce7a5`
- Verdict: `PASS`
- Blocking findings: none
- Architecture conflicts: none
- Semantic delta since review: none

Nonblocking findings were classified before any edit. Duplicated frozen protocol/schedule constants are optional cleanup: current self-tests compare the analysis pins with the benchmark constants and generated schedule bytes. Defensive post-validation branches remain intentionally fail-closed. No correction push or second review was needed.

Final local evidence on the reviewed head: fresh readability verification passed; `@all`, `@doc`, formatting, historical-manifest positive/mutation gates, version smoke, whitespace, and clean-tree checks passed; all 847 compiled tests passed. Local sanitizer crams are limited by Codex ptrace, with the direct runtime suite passing; required GitHub `Development gate`, clang/GCC native parity, Governance, and GM12B contexts provide final merge evidence.

Remaining work is explicitly outside this slice: inferential tables, uncertainty/effect sizes, blinded rescoring, published-number provenance, and measured claim gates. Squash auto-merge will be enabled; GitHub will wait for every required check.

PR #109, opened 2026-08-20
HB.2a: add strict host transport codec

jbwinters:
## Outcome

Adds the strict, bounded library codec for the first Jacquard host carrier. It
implements the frozen `stdio-u32-json-v0` framing and handshake contract while
leaving all existing language, evaluator, scheduler, CLI, and native-runtime
behavior unchanged.

This is the transport/selection slice of HB.2. It deliberately does **not**
make a runnable worker or claim that Jacquard can execute hosted effects yet.
Typed descriptors and boundary values come next; the affine effect lifecycle
and opt-in worker CLI follow after that.

## Included

- unsigned 32-bit big-endian frame encoding and bounded stream I/O
- strict UTF-8 JSON-object validation, including recursive duplicate-key,
  depth, collection-size, numeric-form, and trailing-data rejection
- deterministic `core_hello` and exact `host_select` parsing
- positive component-wise limit negotiation and mandatory terminal capacity
- exact shutdown and shutdown-ack envelopes
- stable E1600/E1601/E1602/E1608/E1611 failure classification without
  reflecting refused payload bytes
- 18 focused unit/property cases, including 300 generated frame round trips
- current documentation and successor evidence inventory updated from 882 to
  900 repository tests

## Explicit exclusions

No store lookup, checker/evaluator integration, target or interface preflight,
boundary-value codec, operation exchange, adapter, socket, server, or CLI is
added here. The public carrier remains unavailable until the later HB.2 slices
complete.

## Evidence

- independent Claude Fable 5 review at xhigh: **PASS**, no required code change
- reviewed and final head: `009f152cbcd65758b7ee6dececb72dbcc4890393`
- `dune runtest`: **900/900**, 424.475s, QCheck seed `762720898`
- focused host-protocol codec suite: **18/18**
- `dune build @all`: pass
- `dune build @doc`: pass
- `dune fmt`: clean
- meaningful whitespace check: clean

The review recorded two forward-looking worker concerns: dispatch decoded
message kind before body parsing to preserve state diagnostics, and specify a
fallback for diagnostics larger than a selected frame. Both are intentionally
owned by the later worker-state-machine slice and do not change this frozen
codec contract.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -118,10 +118,11 @@ When adding valid corpus files, regenerate the golden hashes with
 - Handlers and evaluation: `src/eval.ml`, `test/test_handlers.ml`,
   `test/test_gauntlet_handlers.ml`.
 - Host integration contracts: `docs/host-boundary.md` and
-  `spec/host-protocol-v0.md`. HB.1 freezes envelopes and schema/state vectors,
-  but no executable carrier ships yet. Existing evaluator capture and
-  host-readiness modules are internal seams, not a public ABI; adapters and
-  application servers belong outside this repository.
+  `spec/host-protocol-v0.md`; strict framing and selection codec:
+  `src/host_protocol_v0.ml`, `test/test_host_protocol_codec.ml`. No runnable
+  worker ships yet. Existing evaluator capture and host-readiness modules are
+  internal seams, not a public ABI; adapters and application servers belong
+  outside this repository.
 - Dist and inference: `prelude/06-dist.jqd`, `prelude/13-dist-lib.jqd`,
   `src/infer_dist.ml`, `test/test_infer.ml`.
 - Warp: `prelude/15-warp.jqd`, `prelude/16-gen.jqd`, `src/warp.ml`,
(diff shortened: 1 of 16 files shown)


jbwinters:
## Independent review and exact-head readiness

Reviewed/final head: `009f152cbcd65758b7ee6dececb72dbcc4890393`

Claude Fable 5 at xhigh returned **PASS** with no required code change. The
review directly checked framing bounds, Unicode-scalar UTF-8 handling,
recursive duplicate-key and structure limits, exact hello/selection/shutdown
envelopes, mandatory terminal capacity, hostile-vector agreement, fixed
diagnostics, allocation order, and isolation from the language core.

Finding classification:

- current blockers: none
- semantic corrections: none
- deferred worker debt: dispatch `kind` before choosing a body parser for
  cross-state E1608 behavior; define the encode-failure fallback for a
  diagnostic larger than a negotiated frame
- false assumptions: none
- prose/format-only observations: one cosmetic truncated-prefix cause string

The only condition on the review PASS was an exact-head full run after the
last test strengthening. That condition is now satisfied:

- repository suite: **900/900**, 424.475s, seed `762720898`
- focused codec suite: **18/18**
- build-all, documentation build, formatting, and meaningful whitespace
  checks: pass
- worktree: clean; reviewed and tested head unchanged

